PR #70, opened 2020-11-02
Develop database 

### LINKED ISSUE #51: Setup Database
As a dev, I need a place to store all the system and sensor information in a secure database.

Setup an internal database and connect it into the application.

Acceptance Criteria:
- [x] Setup internal database
- [x] Document which database is implemented and why

Notes;
H2, Derby, HSQLDB are possible solutions.

### LINKED ISSUE #65: Implement Database Design
As a developer, I want to implement the database design for real world use.

Acceptance Criteria:
- [x] The database design is implemented as what is described in the database documentation.

najimgh:
Embedded in-memory H2 database setup and tested #51

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -6,8 +6,8 @@
         <sourceOutputDir name="target/generated-sources/annotations" />
         <sourceTestOutputDir name="target/generated-test-sources/test-annotations" />
         <outputRelativeToContentRoot value="true" />
-        <module name="Capstone" />
         <module name="notorious-eng" />
+        <module name="Capstone" />
         <module name="cbms" />
       </profile>
     </annotationProcessing>
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -8,7 +8,7 @@
       </list>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="15" project-jdk-type="JavaSDK">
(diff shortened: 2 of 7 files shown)


PaulMicu (changes requested):
Do not put your code in the main object, create a db object in /src/main/java/com/cbms/source/local/
From our conversation yesterday i was under the impression that the db will be encapsulated in a jar file, however i dont see it in your pull request

najimgh:
According to Jeremy, it will not be necessary to encapsulate the db in a jar file. He will explain more in detail soon.

PaulMicu:
@JayChouteau Please explain how data is persistent from launch to launch of the application

JayChouteau:
> @JayChouteau Please explain how data is persistent from launch to launch of the application

My understanding from talks with the whole team is that we didn't want something persistent. That is the whole reason we went for a selection between 3 in-memory  dbms types, meaning the db would reside in the main memory and not the hard disk, resulting in its deletion after the application closes. If that wasn't our intention, why did we target those 3 dbms? We understood that we need the db to be embedded. A lot of dbs can be embedded, they almost all have embedded versions. If we wanted embedded and persistent, we would have targeted other dbms, instead of just picking one of those 3.

Now, if persistency is what we want, then it seems there are ways to make h2 persistent, but we will need to make some modifications.

PaulMicu (changes requested):
please include the scripts in this pull request

PaulMicu (approved):

PR #69, opened 2020-11-02
28 29 30 data pre processing

PaulMicu:
#28 and #30 are included in this pull request, as well as a first draft of the model #8 that we used for testing purposes. For issue #29 we are still blocked by the lack of a database

Issue #28, opened 2020-09-23
Clean the data


Shirwarsame:
As a developer, I want to implement the data preprocessing mechanism so that we have clean data

Acceptance Criteria:
- [x] Sensor noise has been reduced to an acceptable threshold within the dataset.

Khal-Shah:
- [ ] Perform statistical summaries and remove instances that aren't significant.

PR #82, opened 2020-11-06
28 30 data cleaning enhancements

Khal-Shah:
#28 
#69 

The original method of applying the AttributeSelection filter from the Weka Library to clean the data potentially removes attributes that can hold significant information that we might want to keep (Ex: Sensor 9, 14).

This additional method ensures _only_ the attributes that have no fluctuation at all (standard deviation = 0) are removed to produce a minimally reduced dataset. Unlike the first method, this doesn't check for any correlation with the RUL, but rather the values of the attribute themselves (if a column has no fluctuation, attribute corresponding to that column is removed). 

Either of these methods can be used to clean the data and then observe which one creates a more accurate model. 

Refactoring: deletes "train_FD001_filtered.arff" file that did not have correct data.


PaulMicu (changes requested):
can you please change the name of the function to something more descriptive, thanks

Khal-Shah:
> can you please change the name of the function to something more descriptive, thanks

Do you mean the process function? 
I named it process2 since the first one there was just called process. 

Should I rename both or just process2()?

PaulMicu:
if there was only one function process would be fine but since we have two that do a similar thing i think we should call it processFullReduction and processMinimalReduction . Or something like that

Khal-Shah:
> if there was only one function process would be fine but since we have two that do a similar thing i think we should call it processFullReduction and processMinimalReduction . Or something like that

Agreed.

Refactoring activity:
Data cleaning method names have been changed:
process() -> processFullReduction()
process2() -> processMinimalReduction()

PaulMicu (commented):
thank you for the correction

PaulMicu (approved):

Issue #46, opened 2020-09-27
Display System Dashboard Details (Thumbnail View)


Shirwarsame:
As a user I want to see main system details for all systems.

Acceptance Criteria:
- [ ] System Type
- [ ] Serial Number
- [ ] RUL
- [ ] RUL Model
- [ ] System Health (Dummy data)

jeffbuonamici:
Make sure it follows current UI Prototype:

https://notorious-eng.atlassian.net/wiki/spaces/DOCUMENTAT/pages/24379414/UI+Prototypes

jeffbuonamici:
Version 0.6 is currently the latest update:
https://notorious-eng.atlassian.net/wiki/spaces/DOCUMENTAT/pages/87392587/v0.6

Issue #44, opened 2020-09-27
Display Selected System Information


Khal-Shah:
As a user, I want to view the detailed system information for the system that I selected.

Acceptance Criteria (the following items should be displayed):
- [ ] RUL
- [ ] Sensor information
- [ ] RUL model
- [ ] System Type
- [ ] System Name
- [ ] System Serial Number
- [ ] System Description
- [ ] Manufacturer
- [ ] Site Location
- [ ] Add(Or Select) Model Button

jeffbuonamici:
Make sure it follows current UI Prototype:

https://notorious-eng.atlassian.net/wiki/spaces/DOCUMENTAT/pages/24379414/UI+Prototypes

jeffbuonamici:
Version 0.6 is currently the latest update:
https://notorious-eng.atlassian.net/wiki/spaces/DOCUMENTAT/pages/87392587/v0.6

PR #85, opened 2020-11-07
44 46 UI Details

jeffbuonamici:
#44 #46

- Created controllers for each scene
- Created CSS files for each scene
- Implemented the necessary JavaFX elements to display the data.
- Created Engine and Sensor classes to hold data in a usable format for the front end.
- Created the onClick events to link between scenes.


PaulMicu (changes requested):
please remove target files from pull request

src/main/java/com/cbms/app/Engine.java
```diff
@@ -0,0 +1,101 @@
+package com.cbms.app;
+
+public class Engine {
```
rdsaliba: We created another model for this called "Asset"

rdsaliba (commented):
I believe you must switch your engine class to our asset class to match the mimosa standard. Let me know what you think.

jeffbuonamici:
I believe I fixed your requests.
- I'm now using Asset and AssetAttribute objects instead of Engine and Sensor
- Deleted target folder files that were committed

Also, Paul helped me update the branch from the Backend Connections branch to avoid merge errors I believe.

rdsaliba (approved):
Ready to merge

Issue #98, opened 2020-11-11
Database Enhancement


JayChouteau:
Modifications to the DB to accommodate the new requirements

Acceptance criteria:
- [x] Adding an asset type table
- [x] Adding Model parameters (boundaries) to the asset type table
- [x] Associate the asset type with model

PaulMicu:
Added a new Asset type parameters table to allow the user to drive the number of parameters used for the recomendation

Issue #48, opened 2020-09-27
(F11) Remove a system type


rdsaliba:
As a user, I want to remove a system type and all related assets from the application,

Acceptance Criteria:
- [x] It is possible to delete/remove a system type from the application through the UI
- [x] All assets of that type have been deleted: they are no longer displayed on the app
- [x] Confirm dialog for deleting pops up when deleting an asset type
- [x] Removing a system type should be done through the "System type Info" view in the UI
- [x] Unit tests

PO question:
Issue #48
- What do we mean by removing a system type? Do we mean that the user can delete a system type? 
And will this lead to all the related systems to be deleted as well or having them simply without a type? Or do we mean that
the system type and all related systems will be set as inactive/unaccounted for in the display/calculations?

- What about editing a system type? Is that something we would want to do?

PO answer (PO meeting on 16-11-2020):
- removing a system type means that the system type and all assets of that type in the system will be deleted
- The only edit we want to do on a system type could be to rename it -> updating this change on all assets of that type

PR #99, opened 2020-11-11
adding new tables

PaulMicu:
As described in #98 I added 2 new tables for the asset_type and the trained_model. The asset_type table also holds boundaries for the assessment calculation

this script needs to be run after the v3 script you can find in the corresponding Zip in the script folder

here is an updated visual interpretation of the database to help review the pull request
V4 01


src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
...
+Adding one asset type for engin with dummy values
+*/
+INSERT INTO asset_type(name, boundary_replace, boundary_repair) VALUES("Engine","99.8","55.5");
+
+/* 
+Adding the asset type id "1" to all current asset
+*/
+UPDATE asset SET asset_type_id = 1 WHERE asset_id>0;
+
+select * from asset
```
JayChouteau: The statement should end with an end of the line ";" symbol

src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
...
+ENGINE = InnoDB;
+SET FOREIGN_KEY_CHECKS=1;
+
+/*
+Adding one asset type for engin with dummy values
+*/
+INSERT INTO asset_type(name, boundary_replace, boundary_repair) VALUES("Engine","99.8","55.5");
+
+/* 
+Adding the asset type id "1" to all current asset
```
JayChouteau, line 102: Maybe leaving a TODO for adding system types desired names would be a good idea to remember to come back to it. This coding action isn't part of the scope of this PR and needs to come from the PO, but a TODO would help us remember once we get that information.

src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
...
+    FOREIGN KEY (`asset_type_id`)
+    REFERENCES `cbms`.`asset_type` (`asset_type_id`),
+  CONSTRAINT `fk_table1_dataset1`
+    FOREIGN KEY (`dataset_id`)
+    REFERENCES `cbms`.`dataset` (`dataset_id`))
+ENGINE = InnoDB;
+SET FOREIGN_KEY_CHECKS=1;
+
+/*
+Adding one asset type for engin with dummy values
```
JayChouteau: engin -> engine

src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
...
+  `asset_type_id` INT(11) NOT NULL AUTO_INCREMENT,
+  `name` VARCHAR(150) NOT NULL,
+  `boundary_replace` VARCHAR(45) NULL,
+  `boundary_repair` VARCHAR(45) NULL,
+  PRIMARY KEY (`asset_type_id`))
+ENGINE = InnoDB;
+
+-- -----------------------------------------------------
+-- alter Table `cbms`.`asset`
+-- removing type colomn, addning asset_type_id colomn and adding the association with Asset_type table
```
JayChouteau: colomn -> column

src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
...
+@Autor Paul Micu
+@version 4.01 (sprint 4, script 1)
+*/
+
+SET FOREIGN_KEY_CHECKS=0;
+drop table if exists asset_type;
+-- -----------------------------------------------------
+-- Table `cbms`.`asset_type`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `cbms`.`asset_type` (
```
JayChouteau, line 29: Implementation well done with the primary and foreign keys

src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
+/**
+
+this script will add the asset_type table and make the association with asset
+the script will also add a value in the type and set all current asset's types to this new one
```
JayChouteau: The script will also add an entry for this new asset type table and set all current asset's types (in the asset table) to this new asset type entry

src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
...
+
+this script will also create a table to store the trained models
+the table will take the modele id, the asset type id and the dataset id as primaries keys
```
JayChouteau: Instead of explaining HOW we are doing the table, I think it would be more relevant to explain WHAT the table is doing, i.e.: what is a trained model for and what is it holding.

src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
...
+it also has a colomn for storing the path to the model once we encapsulate that in a future issue
```
JayChouteau: colomn -> column

src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
...
+-- removing type colomn, addning asset_type_id colomn and adding the association with Asset_type table
+-- -----------------------------------------------------
+ALTER TABLE asset DROP COLUMN type;
+ALTER TABLE asset ADD asset_type_id int;
+ALTER TABLE asset MODIFY asset_type_id int AFTER name;
+create INDEX `fk_asset_asset_type1_idx` on asset (`asset_type_id` ASC);
+ALTER TABLE asset
+  add CONSTRAINT `fk_asset_asset_type1`
+    FOREIGN KEY (`asset_type_id`)
+    REFERENCES `cbms`.`asset_type` (`asset_type_id`);
```
JayChouteau, line 49: Adding an empty line after this one would help separate the logic blocks between asset alteration and trained model table dropping/creation

src/main/resources/script/cbms_scriptsV4/V4.01.sql
```diff
@@ -0,0 +1,73 @@
...
+drop table if exists trained_model;
+-- -----------------------------------------------------
+-- Table `cbms`.`trained_model`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `cbms`.`trained_model` (
+  `model_id` INT(11) NOT NULL,
+  `asset_type_id` INT(11) NOT NULL,
+  `dataset_id` INT(11) NOT NULL,
+  `location` VARCHAR(255) NULL,
```
JayChouteau: Since other entities, such as assets, also have an attribute called "location" that refers to a physical location, this name might lead to confusion. Would be more relevant to call it "path" or "path_location" or else.

JayChouteau (changes requested):
Logic and coding are structured, clear, and complete. There are some parts that will refinement depending on the information/direction the PO gives us, so adding a few TODOs would help us remember to check on him. I left some review changes mainly on comments, possible todos and on 1 naming issue. Very good work.

JayChouteau:
@PaulMicu 
The asset type table doesn't seem flexible in the sense that it doesn't allow to add more recommendations possibilities and values. This comment is opened to design discussion and was originally suggested by @rdsaliba (and I agreed with him). In our design, I feel like we should let the user determine the number of recommendations possible for a system type along with the name/display of the recommendation and the lower bound associated with it. 
-> An example of this design would be to have a table "status" relating to "asset_type". The "status" table would have asset_type_id as a foreign key reference, status_id, status name/display, and a lower boundary as attributes (columns). The asset_type would hold the asset_type_id and name attributes.

PaulMicu:
@JayChouteau I see what you mean. It's not a bad idea, it will make the changes to make it a little more versatile. I'll implement this new criteria asap

src/main/resources/script/cbms_scriptsV4/V4.01.sql
PaulMicu: this will be part of issue #48, for now it needs something in the table for the assets we already have, we can change the name of the current type of assets at anytime to whatever the PO wants

PaulMicu: This will be addressed in issue #48. The name of the asset type can be changed at any time

src/main/resources/script/cbms_scriptsV4/V4.01.sql
PaulMicu: I thought it was pretty self-explanatory but I can add some more details

PaulMicu:
After the latest changes the database model looks like this : 
V4 01

rdsaliba (approved):
Thank you for making the adjustments @PaulMicu. We'll see how this works out in the future.

JayChouteau (approved):

JayChouteau:
@PaulMicu Good job for proceeding with the requested changes. I think this new design makes it a lot easier to accept different kinds of input and nice modifications to the rest of the changes. We are good to go.
Small question:
- It seems we are merging into master, shouldn't we merge into develop?

PR #104, opened 2020-11-19
Bug fix/new asset type needed changes in queries

PaulMicu:
After the changes in issue #98, some queries need to be adjusted because the column was renamed from asset.type to asset.asset_type_id


rdsaliba (commented):
Did you check if it was referenced anywhere in the code? Otherwise it's good to go.

PaulMicu:
it was referenced by assetType which would work fine but i renamed it to assetTypeID to be more clear

rdsaliba (approved):

PR #114, opened 2020-11-22
Database connection bug

najimgh:
- Added functions to stop the database.
- Functions added in Database, DatabaseConnection and ModelController objects


PaulMicu:
There are some conflicts you have to fix before merging, and your changes on the 4.01 script are still present, as discussed please remove 4.01.sql from commit

JayChouteau (approved):
Good job for fixing the bug that was reported on sonarQube. This does take care of closing the database when not needed. I reverted some of the weird formatting changes that were made to the scripts v4.01

PR #115, opened 2020-11-23
Us 94/refactoring activities

rdsaliba:
Inspection fixes
- Code Style issues
- Compiler issues
- Declaration redundancy
- Java language level migration aids
- Javadoc
- Probable bugs
- Proofreading

There's more that needs to be fixed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,8 @@
 target/*
 .idea/compiler.xml
 .idea/misc.xml
+.idea/sonarIssues.xml
+
+cbms.iml
+
+.DS_Store
--- a/src/main/java/com/cbms/preprocessing/DataPrePreprocessorController.java
+++ b/src/main/java/com/cbms/preprocessing/DataPrePreprocessorController.java
@@ -4,8 +4,8 @@
   it will return an Instances object with the attributes and their corresponding data selected by the algorithm removed
 
   @author Paul Micu
- * @version 1.0
- * @last_edit 11/01/2020
+  @version 1.0
+  @last_edit 11/01/2020
  */
 package com.cbms.preprocessing;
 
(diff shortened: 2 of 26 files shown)


src/main/java/com/cbms/preprocessing/DataPrePreprocessorController.java
```diff
@@ -1,31 +1,26 @@
...
- * @author Paul Micu
+/*
+  This Controller will handle the preprocessing of the data.
+  it requires the dataset to reduce in an Instances object
+  it will return an Instances object with the attributes and their corresponding data selected by the algorithm removed
+
+  @author Paul Micu
  * @version 1.0
  * @last_edit 11/01/2020
  */
```
PaulMicu: Any reason why you only removed some of the '*' in this file and the others(i'm not gonna leave a comment on them because there's a few)

PaulMicu:
Are the changes in these files necessary?
.idea/sonarIssues.xml
cbms.iml

.gitignore
```diff
@@ -23,3 +23,8 @@
 target/*
 .idea/compiler.xml
 .idea/misc.xml
+.idea/sonarIssues.xml
```
rdsaliba, line 26: @PaulMicu 
Hi Paul, as you can see I tried to exclude them in the gitignore but it has failed me every single time...

PaulMicu (changes requested):

src/main/java/com/cbms/preprocessing/DataPrePreprocessorController.java
rdsaliba: This was due to automatic fix on intellij, I'll fix those and push the changes. Good catch.

PaulMicu (approved):

Issue #113, opened 2020-11-22
Refactoring - Analyze impact of issue #112 on code and adjust it


JayChouteau:
This issue is to update the source code following the updated database design in issue #112. 

Changes will affect:
- [x] database java file
- [x] UI controller files

Issue #112, opened 2020-11-22
Refactoring - Datasets concept removal


JayChouteau:
The concept of the dataset has to be removed from our database design. This includes:
- [x] dataset table
- [x] operational_conditions table
- [x] Asset_dataset association table
- [x] Dataset_condition association table

In the trained_model table, remove:
- [x] Dataset_id

The reason is that the PO doesn't want different trained models for the same asset type. Datasets shouldn't have an influence on the model, only the asset type should.

Operational conditions and test_or_train value needs to be added somewhere else:
- [x] Add a column (attribute) in the asset table to identify which assets are for training
- [x] Convert operational conditions into an attribute

PR #116, opened 2020-11-26
#112 removing dataset concept

PaulMicu:
This PR is for the script only, adjusting the code will be done with issue #113

### DIFF
--- a/src/main/resources/script/cbms_scriptsV4/V4.02.sql
+++ b/src/main/resources/script/cbms_scriptsV4/V4.02.sql
@@ -1,0 +1,87 @@
+-- This script will make the changes AS requested in issue #112
+
+-- Adding the Archived column in asset to replace the train from dataset and transfer values
+ALTER TABLE cbms.asset ADD COLUMN archived INT(1) NOT NULL DEFAULT '0';
+UPDATE cbms.asset a, cbms.dataset_asset_assoc daa, cbms.dataset d SET a.archived='1' WHERE 
+a.asset_id = daa.asset_id AND
+daa.dataset_id = d.dataset_id AND
+d.train = '1';
+
+-- Remove all constraints, drop the primary key, remake constraints, remake primary key without dataset
+-- Remove dataset_id column from table trained model
+ALTER TABLE cbms.trained_model ENGINE=InnoDB;
+ALTER TABLE cbms.trained_model DROP FOREIGN KEY fk_table1_dataset1;
+ALTER TABLE cbms.trained_model DROP FOREIGN KEY fk_table1_asset_type1;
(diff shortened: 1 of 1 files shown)


jeffbuonamici:
It looks alright to me, but can someone with better SQL knowledge take a look at this to triple check.

jeffbuonamici (approved):

src/main/resources/script/cbms_scriptsV4/V4.02.sql
```diff
@@ -0,0 +1,87 @@
...
+d.train = '1';
+
+-- Remove all constraints, drop the primary key, remake constraints, remake primary key without dataset
+-- Remove dataset_id column from table trained model
+ALTER TABLE cbms.trained_model ENGINE=InnoDB;
+ALTER TABLE cbms.trained_model DROP FOREIGN KEY fk_table1_dataset1;
+ALTER TABLE cbms.trained_model DROP FOREIGN KEY fk_table1_asset_type1;
+ALTER TABLE cbms.trained_model DROP FOREIGN KEY fk_table1_model1;
+DROP INDEX fk_table1_dataset1_idx ON cbms.trained_model;
+DROP INDEX PRIMARY ON cbms.trained_model;
```
JayChouteau, line 17: This line of code works on MariaDB, but doesn't work on SQL server. The alternative syntax on SQL server (if we want everybody in the team to run it), would be:
ALTER TABLE cbms.trained_model DROP primary key;

JayChouteau (changes requested):
Overall, very good work to remove the concept of datasets and shift towards a focus around the asset type as required by the client. I ran the script and it works as intended with all the tables correctly updated. (one exception - see review comment). I like this idea of keeping the operational conditions saved as attributes that can be applied to certain measurements/assets. However, applying them now to the attribute measurements might have been a bit overkill as the client said we didn't care about them and it adds a lot of unnecessary information and entries into our attribute measurement tables. That being said, the cost in time isn't expensive at all and the information is already there in case of future needs, which would seem logical given the motive of use for this project,  so leaving this adding of information is fine. Very structured and complete.

src/main/resources/script/cbms_scriptsV4/V4.02.sql
PaulMicu: I know In doesn't work on MySQL but since we're using MariaDB I don't see the issue. If anyone is using the wrong database it's up to them to correct that and use the chosen one.

JayChouteau: I agree with that statement. I just wanted to make sure. Nice job, I'll approve then!

JayChouteau (approved):

PR #118, opened 2020-11-28
#41 remove asset

### LINKED ISSUE #41: Remove an asset through the UI
As a User, I want to delete an asset from the system so that it is removed from the monitored list.

Refer to UI Prototypes: https://notorious-eng.atlassian.net/l/c/Y5Fr7SZ5

Acceptance Criteria:
- [x] Take the input from the UI and remove the asset from the DB
- [x] Prompts the user with a message "Are you sure you want to permanently delete this item?" before deletion
- [x] Make sure the systems view page is refreshed (asset is deleted from the page - UI)

jeffbuonamici:
This was basically just adding a mouse event handler for the delete button, and then writing an SQL query to delete the asset from the database.

Edit: After talking with @PaulMicu, there will need to be a slight change to the SQL query once his pull request goes through.

### DIFF
--- a/src/main/java/com/cbms/source/local/Database.java
+++ b/src/main/java/com/cbms/source/local/Database.java
@@ -279,6 +279,13 @@ public class Database {
      * @param id
      */
     public void deleteAssetByID(int id) {
-
+        String query = "DELETE FROM attribute_measurements WHERE asset_id = " + id;
+        executeQuery(query);
+        query = "DELETE FROM asset_model_calculation WHERE asset_id = " + id;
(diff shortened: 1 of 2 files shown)


rdsaliba:
@jeffbuonamici if there's a modification to a query required, why don't you include the change within this pull request?

jeffbuonamici:
Because then the code would be merged with an error in it.

JayChouteau:
@jeffbuonamici It seems this branch has conflicts that need to be resolved. Could you take care of it so we can review and merge?

PR #130, opened 2020-12-05
Update issue templates

rdsaliba:


JayChouteau (approved):
This is a good template for bug reports! It will also us to have better descriptions of what is happening, the behavior we expect, and how to reproduce it. Assignees will then be able to leave comments on the resolution in the comments. Nice!

PR #131, opened 2020-12-06
Updating Queries to accomodate the new database design

### LINKED ISSUE #113: Refactoring - Analyze impact of issue #112 on code and adjust it

JayChouteau:
The database design was changed in issue #112 following some discussions with our PO. The conclusions of this discussion, that can be witnessed in the mentioned issue are the following:
- Removal of the dataset concept: we do not care to which dataset the assets, attributes or measurements belong. 
- This also means that we do not care about the difference in operational conditions that are applied to different datasets and how they could have a different impact on the RUL or measurements on the assets. 
- The only thing that will influence the RUL calculation or grouping of data (assets) is the asset type
- The classifiers (trained models) will be based on the asset type and not the dataset
- In our application, we only have 1 asset type, known as 'engine'.

With those conclusions and the update of the database design, some changes were needed to be made on the queries that extract information for the database. Whether it is to create objects or display those objects in the UI, the modifications to the queries and code affected 2 files:
- database.java (where we find all the queries on the db)
- modelcontroller.java (the main controller of our application)

Result:
Starting the application results in a longer start-up process as the preprocessing is done on 4 datasets (1 asset type) now rather than one. The application now also displays all assets (from the 4 datasets test) in the "Systems-dashboard" view as well as the "Systems-list" view instead of only the ones from dataset FD001. The rest of the application behaves normally. 

Eventual todos:
- Add SN values for assets as assets that are over ID=200 do not have SN attached to them and the display of the SN is empty (scripts). 
- Add a dropdown in the "Systems" views to choose which asset type to display assets from.
- Show the asset type name in the "Systems" view on each asset instead of the asset type ID, it would make more sense for a user.

### DIFF
--- a/src/main/java/com/cbms/app/ModelController.java
+++ b/src/main/java/com/cbms/app/ModelController.java
@@ -17,7 +17,6 @@ import com.cbms.source.local.Database;
 import weka.classifiers.Classifier;
 import weka.core.Instances;
 
-import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Map;
 import java.util.TreeMap;
@@ -35,7 +34,7 @@ public class ModelController {
 
     private ModelController() throws Exception {
         db = new Database();
-        //trainingSets = db.getTrainDatasets();
+        trainingSets = db.getTrainAssetTypes();
         instancesSets = new TreeMap<>();
         reducedInstancesSets = new TreeMap<>();
         classifierSets = new TreeMap<>();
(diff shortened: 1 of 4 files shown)


jeffbuonamici (approved):
I went through everything and it looks good. 

I don't know if it's a coincidence, but the application seems to be running slower. This is probably something on my end though.

jeffbuonamici:
Fixed a small bug that caused a crash. The ID of the cancel button on the AddSystems page was incorrectly named.

PaulMicu (approved):
The changes are good, we will touch on this again when refactoring sprint 4-5

Issue #97, opened 2020-11-11
Research - Model Parameters (LSTM)


JayChouteau:
As a user, I would like to change the value of certain LSTM model parameters to tune the model if it is possible.

Acceptance criteria:
- [x] Research on the possibility of editing parameters that are input in the RUL LSTM model
- [x] Assess the impact of such changes on the model calculation and the limitations of it
- [x] Document on Confluence

 Description:
The PO added this requirement with the idea in mind that certain parameters used in the RUL calculation with RUL models could be modified to tune the RUL values. He thought it would be interesting for the user to edit the values of those parameters to see the influence on the model calculation. In this particular case, it would be to check if the LSTMpossesses any parameter as an input that could be tuned within a certain range by a user. An impact assessment of such modification should be performed as well.

Khal-Shah:
Confluence document:
https://notorious-eng.atlassian.net/l/c/kQPfvaS8

Issue #12, opened 2020-09-22
LSTM-Evaluate the model 


Shirwarsame:
As a dev, I want to evaluate the LSTM model's accuracy so that we can compare the calculation of the RUL prediction vs real-world.

Acceptance criteria:
- [ ]

PR #133, opened 2020-12-19
132_DL4J_Integration_LSTM

### LINKED ISSUE #10: LSTM - Create a network model
As a dev, I want to develop a Long Short-Term Memory network model so that it calculates the RUL of an asset.

Acceptance
[ ] Implement new model strategy

### LINKED ISSUE #132: LSTM - Dl4j integration
As a developer, I want to have the right framework to use deep learning with weka.

Acceptance criteria:
- [ ] Have an LSTM Network
- [ ] Have all the needed dependencies for Dl4j integration with our project and weka
- [ ] Have the LSTM Network run successfully

This issue is to be paired with issue #10.

Khal-Shah:
This integrates Weka's DL4J plugin into the project for the LSTM network built in #10 to work. 

Weka's DL4J doesn't have a central Maven repository so the jar file had to be added locally alongside other dependencies. 

The Deep Learning Neural Network takes about ~2~ 15 minutes to train after the refactoring (all datasets).

The network and LSTM model itself will probably need tinkering depending on how accurate it is ( #12) and as our understanding of LSTM is further improved with #97.

**~~Testing out LSTM~~**
~~To run the app using the values generated by LSTM instead of Linear Regression, go on _ModelController_ class in the _app_ package and uncomment line 45 while commenting out line 42. Though the GUI will still say "Linear Regression", the RUL values themselves will be generated using LSTM~~. 
N/A after refactoring.

**Refactoring activities**
The Weka version had to be updated to the latest 3.8.4 version (from 3.6.14 previously). With that, it appears the weka.core.Instance class has been deprecated. This caused errors from the Database class, so the usage of _Instance_ class there has been refactored to _DenseInstance_. Everything seems fine in the other files. 

The modifiers of removeInstance method in LinearRegressionModelImpl class were changed to _public static_ so that it can be used easily from the LSTMModelImpl class.

**Future issues**

- Since this is slowing down starting the app considerably and if the training dataset doesn't change, we don't need to train the model every time we run the app. We should look into only training the data when the app is ran for the first time in a system then saving it locally so that it doesn't need to be ran each and every time.

- Similar to Linear Regression, the GUI for LSTM will need to be added


PaulMicu (changes requested):
There are multiple mentions of "dataset" added in the DataPreProcessorImpl. The concept of a dataset has been removed from the project since sprint 4, please review those methods to not have anything regarding dataset

Khal-Shah:
> There are multiple mentions of "dataset" added in the DataPreProcessorImpl. The concept of a dataset has been removed from the project since sprint 4, please review those methods to not have anything regarding dataset

Comments in DataPreProcessorImpl have been edited as per request.

Previously I was using "dataset" to refer to the assetIDs corresponding to what we knew as our 4 datasets since assetIDs aren't continous.
Ex: 
```
     AssetID 1-100 --> Training dataset1
     AssetID  201-460 --> Training dataset2
     etc
```
     
The DataPreProcessorImpl has been refactored so that it can be used if needed to add RUL column with the correct values for the entire training set (assetID 1-1168).

I also wanted to do it for the testing set, but it appears some of the data is missing in the DB.
Ex: 
```
        Database db = new Database();
        Instances testSet = db.createInstancesFromAssetTypeID(1, 0);

        for(int i = 0; i < 10000; i++)
        {
            System.out.println(testSet.instance(i).toString());
        }
```
The data corresponding to AssetID 103-105 (which corresponds to testing set 1, engine 3-5) are missing as it jumps directly to 106 after 102.

Khal-Shah:
Since this issue really doesn't have anything to do with the preprocessor, the last few commits have been reverted and the DataPreProcessorImpl class is back to its initial untouched state.

A future preprocessor refactoring issue will be more suitable if any changes are needed.

PaulMicu (approved):
good work, thank you for addressing the changes

jeffbuonamici (approved):
Looks good. It's ready to merge to develop.

PR #140, opened 2020-12-27
136 app package refactoring

### LINKED ISSUE #136: Refactoring - App package
This refactoring issue targets the app package but does not cover the item sub-package. This only targets the model controller and the main.

1) Model Controller
- [x] Initializer method modifications needed after review

2) Main
- [x] Initializer method review and modifications needed

PaulMicu:
modified the App package to be more flexible and not calculate the RUL every time it needs to be displayed but only when new measurements are recorded, this is to help speed up the app usage and avoid unnecessary calculations


JayChouteau (approved):
Well done update to change how our app loads the values and updates itself.

jeffbuonamici (approved):
Looks good.

Issue #128, opened 2020-12-03
Testing - Code coverage framework


JayChouteau:
As a developer, I would like to know the code coverage of my unit testing and testing in general.

Acceptance criteria:
- [x] Code coverage can be output from our application (testing part)
- [x] Documentation on Confluence on the fix or setup as well as how to run/output

Description:
Since Jacoco doesn't work (it doesn't provide us with code coverage), we have to find a solution to this issue. We need it to be able to reach our quality metrics goals. There are 2 solutions:
- Trying to fix Jacoco for our application
- Finding another testing coverage framework

PR #147, opened 2020-12-27
#128 Clover Code Coverage Report Generator

Shirwarsame:
Adding the plugins to the pom file to allow for the creation of a coverage report through clover description on Confluence


PaulMicu (changes requested):
the .idea file and the RulTest.Java have nothing to do with this PR, they need to be removed from commit

PaulMicu (approved):

PR #153, opened 2021-01-11
108 Load the database in real-time

### LINKED ISSUE #108: Real-Time: Load the database in real-time
As a user, I want to be able to use the application as a real-time system that takes continuous input.

Acceptance Criteria:
- [x] When the application starts, only measured data up to time cycle 15 is/should have been inserted into the application. This means that the data up to cycle 15 should already be available when starting the application
- [x] Every 5 sec, new data is loaded into the application (database).

PO precision (12-11-2020):
This issue corresponds to a request from the PO to make our application act as a real-time system since this is how it will be used. This issue, loading the database in real-time, represents the continuous action of new data measured by sensors or others being fed to the application in a real-life context.

Description:
Since all we have is existing data, for now, we will load parts of the assets measured data as an initial block and have the rest of the asset measured data from the datasets to be loaded incrementally into the database following a period (based on a time cycle). For example, every 5 sec, data that corresponds to that time cycle frame will be entered into the database for each asset until there is no more entry.

Added information:
The choice of time cycle 15 as a basis for all data is random. It can be modified by the assignees of this task as long as it is pertinent and realistic.
The 5-sec timeframe was randomly chosen. The time chosen can be modified by the assignees of this task as long as it is relevant and realistic. The timeframe has to match the other Real-Time issues.

PaulMicu:
Added the option to turn on real time mode and that will simulate new data coming in every second until all data is read

### DIFF
--- a/src/main/java/com/cbms/app/Main.java
+++ b/src/main/java/com/cbms/app/Main.java
@@ -1,19 +1,38 @@
 package com.cbms.app;
 
 
+import com.cbms.source.local.AssetDAOImpl;
 import javafx.application.Application;
 import javafx.application.Platform;
 import javafx.fxml.FXMLLoader;
 import javafx.scene.Parent;
 import javafx.scene.Scene;
 import javafx.stage.Stage;
-import javafx.stage.StageStyle;
+
+import java.util.Timer;
+import java.util.TimerTask;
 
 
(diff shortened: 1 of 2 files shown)


jeffbuonamici:
It looked good, so I merged.

PR #154, opened 2021-01-11
12_Evaluate_LSTM

### LINKED ISSUE #12: LSTM-Evaluate the model

Khal-Shah:
J Unit test for the LSTM Evaluation method for issue #12.

The previous converted train_FD001_withRUL and test files (needed for testing purposes) were missing the last sensor attribute and have been replaced with the correct files. Also includes the test file that includes only the last instances of trainingSet1 but with the real RUL values.


src/test/java/rul_models/testModel.java
```diff
@@ -1,23 +1,58 @@
...
+        lstmEval.evaluateModel(classifier, testSet);
+        System.out.println(lstmEval.toSummaryString());
+        return lstmEval;
+    }
+
+    @Test
+    public void testLSTMEval() throws Exception
+    {
+        assertTrue("RMSE should be less than 41", LSTMEval().rootMeanSquaredError() < 41);
+    }
```
PaulMicu, line 53: we spoke about this during our meeting, thank you for adjusting it as discussed

PaulMicu (approved):

PR #184, opened 2021-01-17
183 Performance modules decoupling

### LINKED ISSUE #180: Performance - Memory Leak
**Describe the bug**
The application becomes non-responsive. 
Trace: Exception in thread "JavaFX Application Thread" java.lang.OutOfMemoryError: unable to create native thread: possibly out of memory or process/resource limits reached

**To Reproduce**
1. Launch the application
2. Click on "List" view
3. Select any System to view its information
4. Click back on "Systems" on the side bar
5. See error

**Expected behavior**
Clicking on the "Systems" button on the side bar should bring me back to "List" view

**Screenshots**
Screen Shot 2021-01-13 at 3 46 09 PM

 - OS: **macOS Big Sur**
 - Branch: Develop

**Additional context**

### LINKED ISSUE #183: [Performance] Project decomposition
We want to improve the performances of the application by decoupling the different aspects of the project. 
To do this we will create 3 different Modules 
- Server
     this will contain the logic of the application and will run a thread to check and update the rul of assets as necessary 
- JavaFXUserInterface
    This will contain only the JavaFx interface that will display the current asset information
- RealTimeDemo
   This will only contain a simple demo that removes and adds back measurement in the database to simulate new data coming in

This issue was in part to fix the bug #180 and in part to improve the general performance of the application by reducing the number of actif threads and decouple the different aspects of the application.

PaulMicu:
Project restructured into modules.

to run a modules, make sure that the other modules that are a dependency have been compiled and are up to date with the latest changes

this will fix bug #180 and implement  issue #183


JayChouteau (approved):
This is a big project re-decomposition, but it definitely increases the project performance. The app stopped freezing so much and is a lot smoother now thanks to the logic done apart from the ui. Great refactoring that also provides a better example of how our system would behave in a real world application for the real-time part of it with the data being loaded gradually into the database instead of time cycles being read (data already entered). The decouping benefits our application without damaging the maintainability or expandibiility of the software. These changes will need a little adapation from the team, but are welcomed. Approved!

.idea/jarRepositories.xml
```diff
@@ -9,17 +9,37 @@
     <remote-repository>
       <option name="id" value="local-maven-repo" />
       <option name="name" value="local-maven-repo" />
-      <option name="url" value="file:////local-maven-repo" />
+      <option name="url" value="file:////home/paulmicu/Projects/notorious-eng/JavaFXUserInterface/local-maven-repo" />
```
rdsaliba: Has this been causing an issue?
All your paths are specific to your machine, they should be relative.

JavaFXUserInterface/src/main/java/CBMSApplication.java
```diff
@@ -0,0 +1,35 @@
...
+    public static void main(String[] args) {
+        launch(args);
+    }
+
+        @Override
+        public void start(Stage primaryStage) throws Exception {
+            Parent root = FXMLLoader.load(getClass().getResource("/Systems.fxml"));
+            Scene sample = new Scene(root);
+            primaryStage.setTitle("CBMS");
+           // primaryStage.initStyle(StageStyle.UNDECORATED);
```
rdsaliba, line 18: Commented code should be removed

Server/src/main/java/local/AssetDAO.java
```diff
@@ -5,18 +5,17 @@
...
+import app.TrainedModel;
+import app.item.Asset;
 
 import java.util.ArrayList;
 
 public interface AssetDAO {
     ArrayList<Asset> getAssetsToUpdate();
 
-    Asset getAssetToUpdate(int assetID);
+    //Asset getAssetToUpdate(int assetID);
```
rdsaliba, line 18: Is this still useful as well? Dead code should always be removed

.idea/jarRepositories.xml
PaulMicu: Looks like an .idea file was committed accidentally. I'll remove it from the coming

JavaFXUserInterface/src/main/java/CBMSApplication.java
PaulMicu: Code cleanup is out of scope of the pr

Server/src/main/java/local/AssetDAO.java
PaulMicu: Code cleanup is out of scope of this pr

rdsaliba (approved):
Hi @PaulMicu, I have tested your branch on my old machine and this is working way better than it used to. I didn't get the memory leak issue and the system performance was much more snappier (good job on that!).
I took a look on the updated files, the changes are ok.

PR #185, opened 2021-01-18
11 add asset v2

### LINKED ISSUE #11: Add assets through UI
As a user, I want to add assets to the system using HMI so that it reflects the assets to be monitored.

Refer to UI Prototypes: https://notorious-eng.atlassian.net/l/c/Y5Fr7SZ5

Acceptance criteria:
- [x] Get the data from the UI and add it to the DB
- [x] Associate the new entry with the right dataset (op cond, asset type)
- [ ] unit tests

jeffbuonamici:
You can now add assets to the database. However, our application doesn't handle assets well when they don't have sensor data. This is something we will have to look into.

Has been changed to fit with the new code performance refactor.

### DIFF
--- a/JavaFXUserInterface/src/main/java/AddSystemController.java
+++ b/JavaFXUserInterface/src/main/java/AddSystemController.java
@@ -1,13 +1,18 @@
+import app.item.Asset;
+import app.item.AssetType;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
+import javafx.event.EventHandler;
 import javafx.fxml.FXML;
 import javafx.fxml.Initializable;
-import javafx.scene.control.Button;
-import javafx.scene.control.ChoiceBox;
-import javafx.scene.control.TextArea;
+import javafx.scene.control.*;
+import javafx.scene.input.MouseEvent;
+import javafx.util.StringConverter;
+import local.AssetDAOImpl;
 import local.AssetTypeDAOImpl;
 
 import java.net.URL;
+import java.util.Optional;
 import java.util.ResourceBundle;
 
 public class AddSystemController implements Initializable {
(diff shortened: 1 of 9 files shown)


Server/src/main/java/local/AssetDAOImpl.java
```diff
@@ -213,6 +214,30 @@ public void addRULEstimation(Double estimation, Asset asset, TrainedModel model)
...
+            PreparedStatement ps = getConnection().prepareStatement(INSERT_ASSET);
+            ps.setString(1, asset.getName());
+            ps.setInt(2, Integer.parseInt(asset.getAssetTypeID()));
+            ps.setString(3, asset.getDescription());
+            ps.setString(4, asset.getSerialNo());
+            ps.setString(5, asset.getManufacturer());
+            ps.setString(6, asset.getCategory());
+            ps.setString(7, asset.getSite());
+            ps.setString(8, asset.getLocation());
+            ps.setInt(9, Integer.parseInt(asset.getSerialNo()));
```
PaulMicu: the SN can include letters so it needs to be set as a string

JavaFXUserInterface/src/main/java/AddSystemController.java
```diff
@@ -65,5 +110,102 @@ public void initializeFieldValues() {
...
+
+    /**
+     *  Checks to see if values of the asset are filled.
+     *
+     * @param asset
+     * @return
+     */
+    boolean isAssetEmpty(Asset asset) {
+        if(asset.getName() == "" || asset.getAssetTypeID() == "" || asset.getDescription() == "" ||
+                asset.getSerialNo() == "" || asset.getManufacturer() == "" || asset.getCategory() == "" || asset.getSite() == "" || asset.getLocation() == "")
```
PaulMicu: this does not function as intended, you need to use .equals

PaulMicu (changes requested):

JavaFXUserInterface/src/main/java/AddSystemController.java
PaulMicu: or isEmpty()

PaulMicu (approved):

Issue #88, opened 2020-11-08
Real-Time: Display Assets raw data


rdsaliba:
As a user, I want to visualize the assets' raw data so that I have access to the data used for generating the RUL.

Acceptance Criteria:
- [x] This raw data view can be accessed from the System info page (raw data tab)
- [x] The page will show the raw data (all sensor measurements for different time cycles) for that specific system
- [x] The page will periodically refresh itself as new data is coming in (every sec)
- [ ] Unit tests

PO Precision (12-11-2020):
This issue is part of a series of issues to make our application act as a real-time system. This issue is focused on displaying the raw data of an asset in a real-time fashion. The raw data includes sensor measurements from different time cycles. The real-time system behavior is implemented through the incremental display of the raw data as new data is coming into the application (#108 )

Added information:
The timeframe was chosen to fit with the other real-time issues. The time chosen can be modified by the assignees of this task as long as it is relevant and realistic. The timeframe has to match the other Real-Time issues.

PR #187, opened 2021-01-19
Us110 raw data

### LINKED ISSUE #120: UI - Create a "Raw Data" Screen
One of the user stories requires us to display the raw data for a specific asset. This screen needs to be created first in the UI,

Acceptance Criteria:
- [x] Mock-up update in Confluence (link to be provided in this issue)
- [x]  Create a screen based on the mock-up. Make sure all elements are present.
- [x]  The screen should be accessible from the "System information" screen.
- [ ] Unit tests

rdsaliba:
Create a screen based on the mock-up. 
 The screen is accessible from the "System information" screen.


JavaFXUserInterface/src/main/java/SystemInfoController.java
```diff
@@ -31,6 +31,37 @@
...
+    private static final String S21_COL = "s21";
+    private static final String S22_COL = "s22";
+    private static final String S23_COL = "s23";
+    private static final String S24_COL = "s24";
+    private static final String S25_COL = "s25";
+    private static final String S26_COL = "s26";
+    private final String CYCLE_COL = "Cycle";
+    private final String OP1_COL = "OP1";
+    private final String OP2_COL = "OP2";
+    private final String OP3_COL = "OP3";
```
PaulMicu: This is not dynamic enough, the amount and names of the attributes have to come from the database, not hard coded

JavaFXUserInterface/src/main/java/SystemInfoController.java
```diff
@@ -224,4 +266,61 @@ void deleteDialog(MouseEvent mouseEvent) {
...
+        TableColumn s22 = new TableColumn(S22_COL);
+        TableColumn s23 = new TableColumn(S23_COL);
+        TableColumn s24 = new TableColumn(S24_COL);
+        TableColumn s25 = new TableColumn(S25_COL);
+        TableColumn s26 = new TableColumn(S26_COL);
+
+
+        table.setId("RawDataTable");
+        table.getColumns().addAll(cycle, op1, op2, op3, s1,s2,s3,s4,s5,s6,s7,s8,s9,s10,
+                s11, s12, s13, s14, s15, s16, s17, s18, s19, s20, s21, s22, s23, s24, s25, s26);
```
PaulMicu: same as my other comment, this is not flexible enough, don't hard code the sensors directly into the code but dynamically generate them from the info stored in the db

PaulMicu (changes requested):
Please see comments

JavaFXUserInterface/src/main/java/SystemInfoController.java
```diff
@@ -224,4 +266,61 @@ void deleteDialog(MouseEvent mouseEvent) {
...
+        TableView table = new TableView();
+
+        // When TableRow is clicked, send data to SystemInfo scene.
+        table.setRowFactory(tv -> {
+            TableRow<Asset> row = new TableRow<>();
+            row.setOnMouseClicked(event -> {
+                uiUtilities.changeScene(event, row, "/SystemInfo", row.getItem());
+            });
+            return row;
+        });
```
jeffbuonamici: This code is unnecessary.

JavaFXUserInterface/src/main/java/SystemInfoController.java
jeffbuonamici: You need to loop over the sensor data, and create the table columns dynamically.

jeffbuonamici (changes requested):

JavaFXUserInterface/src/main/java/SystemInfoController.java
```diff
@@ -27,10 +34,42 @@
...
+    private static final String S21_COL = "s21";
+    private static final String S22_COL = "s22";
+    private static final String S23_COL = "s23";
+    private static final String S24_COL = "s24";
+    private static final String S25_COL = "s25";
+    private static final String S26_COL = "s26";
+    private final String CYCLE_COL = "Cycle";
+    private final String OP1_COL = "OP1";
+    private final String OP2_COL = "OP2";
+    private final String OP3_COL = "OP3";
```
PaulMicu: This code is never used and must be removed

PaulMicu (changes requested):
Some code needs to be removed(see comment) and the table is missing the Cycles column so that needs to be added

rdsaliba:
Please see attached image of the system with the recent changes.
Screen Shot 2021-01-21 at 12 20 02 AM

rdsaliba:
The data list is not currently being updated when new measures are coming in. Needs to be fixed yet.

JayChouteau:
@rdsaliba Is the pull request only for issue  #120 or for both #120 and #88 ?? I thought this was for 88 only. This changes the scope of things.

PaulMicu (approved):
**accidentally approved but meant to request changes** the order should be reversed, it makes alot more sense that the info at the top is the most recent

PaulMicu (changes requested):
the order should be reversed, it makes alot more sense that the info at the top is the most recent

PaulMicu:
> @rdsaliba Is the pull request only for issue #120 or for both #120 and #88 ?? I thought this was for 88 only. This changes the scope of things.

I think this is correct, the issue is just to display the latest raw data, the other issue is for making sure the view updates with new measurements

rdsaliba:
@PaulMicu User can select the cycle column to sort by asc/dsc in order to see latest data.
As per the acceptance criteria of the Issue, it is not required to display by latest first. Tho, I agree it could be implemented this way in a future enhancement to the system. Then, maybe also we could add another column for "date"...
Yes this is just #120, I got confused.

PaulMicu:
@rdsaliba  I disagree, just because it's not spelled out word for word in the acceptance criteria does not mean it should not be done. But if other people approve it as is I won't oppose the merge, even if it will have to be changed to display the latest rul first for the real time issue(#88).

rdsaliba:
Interesting fact... The default sorting mechanism is not working as expected on this TableView... I have tried changing the object being passed to a Double and Integer but resulted with the same weird sorting... I think, for now it would be best to accept this pull request and attempt to do the sorting in the issue 88

najimgh (changes requested):
Tabs are displayed well, the raw data display looks good and the relevant information is present.
However, upon clicking on a specific system to display its SystemInformation page, the following error occurs:
(This error does not affect the raw data display, however)
Screen Shot 2021-01-22 at 10 55 00 PM

rdsaliba:
Hi @najimgh, I have tested the application from my side and did not get any error. I believe this must be something unrelated to the above changes ...

najimgh (approved):
After some investigation on the error, it looks like it is unrelated to your PR. Everything else looks good.

PaulMicu:
thank you for implementing the ordering from the most recent to the oldest

PaulMicu (approved):

PR #191, opened 2021-01-20
#156 Populate the "System Type List" screen

### LINKED ISSUE #156: Populate the "System Type List" screen
Since many requirements ask us to add/remove/edit with system types as well as letting the user establish thresholds for the recommendations of that type, having a screen for system types and their different parameters become necessary.

Acceptance Criteria:
- [x]  Add a side menu button to link to this page called "System types" with a different icon than the one for "System"
- [x]  List of system types
- [x]  Name of the system type
- [x]  Number of associated assets
- [x]  Recommendation thresholds
- [x]  Button to add a system type (links to proper page)

Shirwarsame:
Created the System Type List view which displays a table view containing a list of the system types
 including their names, number of associated assets, and their recommendation thresholds.

To test view edit line 15 in CBMSApplication from /Systems.fxml to /SystemTypeList.fxml


JavaFXUserInterface/src/main/java/SystemList.java
```diff
@@ -0,0 +1,82 @@
+import javafx.beans.property.SimpleStringProperty;
+
+public class SystemList {
```
JayChouteau: This naming leads to confusion. It would be more appropriate to call it SystemTypeList or AssetTypeList.

JavaFXUserInterface/src/main/java/SystemTypeController.java
```diff
@@ -1,43 +1,125 @@
...
+    @FXML private TableColumn<SystemList,Integer> columnAssociatedAssets;
+    @FXML private TableColumn<SystemList,Double> columnOk;
+    @FXML private TableColumn<SystemList,Double> columnAdvisory;
+    @FXML private TableColumn<SystemList,Double> columnCaution;
+    @FXML private TableColumn<SystemList,Double> columnWarning;
+    @FXML private TableColumn<SystemList,Double> columnFailed;
+
+    //Configure buttons
+    @FXML private Button systemMenuBtn;
+    @FXML private Button systemListBtn;
```
JayChouteau: Naming convention issue for the buttons, Since both buttons are used on the same panel, they should follow a similar naming convention. For example, you would have "systemMenuBtn" and "systemTypeMenuBtn"

JavaFXUserInterface/src/main/java/SystemTypeController.java
```diff
@@ -1,43 +1,125 @@
...
+        attachEvents();
+        //set up the columns in the table
+        columnName.setCellValueFactory(new PropertyValueFactory<SystemList,String>("name"));
+        columnAssociatedAssets.setCellValueFactory(new PropertyValueFactory<SystemList,Integer>("associated_assets"));
+        columnOk.setCellValueFactory(new PropertyValueFactory<SystemList,Double>("value_ok"));
+        columnAdvisory.setCellValueFactory(new PropertyValueFactory<SystemList,Double>("value_advisory"));
+        columnCaution.setCellValueFactory(new PropertyValueFactory<SystemList,Double>("value_caution"));
+        columnWarning.setCellValueFactory(new PropertyValueFactory<SystemList,Double>("value_warning"));
+        columnFailed.setCellValueFactory(new PropertyValueFactory<SystemList,Double>("value_failed"));
+
```
JayChouteau: The initialize method should only be about invoking other methods and not doing logic for comprehensiveness and readability of the code. Instead of setting the property values of all columns within the initialize method, it would be better to create a method that does all of these and this method gets called in initialize.

JavaFXUserInterface/src/main/java/SystemTypeController.java
```diff
@@ -1,43 +1,125 @@
...
+        //Fill table with database information
+        tableView.setItems(getSystemList());
```
JayChouteau, line 49: This could also be part of the same method as the one mentioned above or another method

JavaFXUserInterface/src/main/java/SystemTypeController.java
```diff
@@ -1,43 +1,125 @@
...
+     * @author Shirwa
+     */
+    private ObservableList<SystemList> getSystemList() {
+        ObservableList<SystemList> systemlist = FXCollections.observableArrayList();
+
+        AssetTypeDAOImpl assetTypeDOA = new AssetTypeDAOImpl();
+
+        ArrayList<AssetType> systemlistname = assetTypeDOA.getAssetTypeList();
+        ArrayList<Integer> systemlistcount = assetTypeDOA.getAssetTypeIdCount();
+
```
JayChouteau, line 64: This is related to the naming confusion mentioned above about system list vs system type list

JavaFXUserInterface/src/main/java/SystemTypeController.java
```diff
@@ -1,43 +1,125 @@
...
+                uiUtilities.changeScene(mouseEvent, "/AddSystemType");
+            }
+        });
+
+        //Attach link to systemListBtn to go to SystemTypeList.fxml
+        systemListBtn.setOnMouseClicked(new EventHandler<MouseEvent>() {
+            @Override
+            public void handle(MouseEvent mouseEvent) {
+                uiUtilities.changeScene(mouseEvent, "/SystemTypeList");
+            }
```
JayChouteau: The 3 above events can be reformatted to be set as lambda expressions. Intellij usually suggests it to you, so check it out. If you need help, just ask.

Server/src/main/java/local/AssetDAOImpl.java
```diff
@@ -34,6 +34,9 @@
     private static final String GET_LATEST_MEASUREMENT_TIME_FROM_ASSET_ID ="SELECT time FROM attribute_measurements am, attribute att WHERE att.attribute_id=am.attribute_id AND am.asset_id = ? order by time desc limit 1";
     private static final String INSERT_ASSET = "INSERT INTO asset (name, asset_type_id, description, sn, manufacturer, category, site, location, unit_nb) values(?,?,?,?,?,?,?,?,?)";
 
+
+
+
```
JayChouteau: Remove the added white space.

Server/src/main/java/local/AssetTypeDAOImpl.java
```diff
@@ -14,6 +14,42 @@
...
+
+    /**
+     * This will return an arraylist of the count for all the asset type ids from the
+     * asset table
+     * @author Shirwa
+     */
+    public ArrayList<Integer> getAssetTypeIdCount() {
+        ArrayList<Integer> assets = new ArrayList<>();
+        try {
+            PreparedStatement ps = getConnection().prepareStatement(GET_ASSET_TYPE_ID_COUNT);
```
JayChouteau: When you use prepared statements, it opens a connection to the database. In order to make sure that this connection gets closed, you either have to use the try-with-resources or use a finally block. The first option is easy to implement and suggested to you by Intellij. You can ask if you need help with this,

Server/src/main/java/local/AssetTypeDAOImpl.java
```diff
@@ -14,6 +14,42 @@
...
+                assets.add(rs.getInt("count"));
+        } catch (SQLException e) {
+            e.printStackTrace();
+        }
+        return assets;
+    }
+
+    public double getAssetTypeBoundary(int asset_type_id, int boundary_type){
+        double boundary = 0.0;
+        try {
```
JayChouteau: Same comment as above. You can see the changes similar to this in the SonarQube refactoring pr.

JayChouteau (changes requested):
Well done on this user story. The creation of a utility class to help you set up the table view is a good idea. It works fine and gets the boundary dynamically from the database. Good use of the current app structure. I left a few change requests, mostly about naming and code handling. Please reach out to me if you have any questions. Also, a few other comments:
1) How do you get to the system type info page? There doesn't seem to be a way to get to it from the system type list page.
2) The table is a bit awkwardly placed on the page. It is right next to the left side panel without space between the panel and the table, but there is a big blank space to the right. Extending the table to the right or centering it might be more pleasing to the eye.

JayChouteau (approved):

JayChouteau:
Thanks for addressing my requests. The code is cleaner now and more comprehensive. Also, the dump file works correctly and the UI looks nice. Nice job, this can get merged.

PR #192, opened 2021-01-20
#182 Negative RULs

### LINKED ISSUE #182: Fix Negative RUL Values
**Describe the bug**
For assets ID# 200+ (asset type ID 2 to 4), when running the application, the estimates for the RUL are negative values (should be > 0). 

**To Reproduce**
1. Run the application
2. On main page, scroll down to Assed IDs 200+
3. The RUL displayed is negative. 

**Expected behavior**
RULs should always be >= 0.

**Screenshots**
image

 - OS: Mac OS, Windows (10), Linux
 - Current version

**Additional context**
Remaining Useful Life should not be negative. 
The preprocessor may not be adding the correct RUL values when calling addRULCol.

- [x] Refactor Preprocessor to check the add RUL column function is adding the correct RUL values for all assets. 
- [x] The RULs displayed should be >= 0 all the time for the whole duration (even during Real time evaluation).

Khal-Shah:
This PR fixes a previous issue where the training instances after the first asset type did not have correct RUL values.
For the negative RULs, running the new dump file (cbms_scriptsV7.zip) which saves the status of the database after retraining all 4 asset type IDs seems to fix the issue. There are still around 20 negative RULs out of 700+ assets, which are likely outliers due to our RMSE.


PaulMicu (changes requested):
you accidently removed some files, please add those files back

PaulMicu:
still not good, you need tto remove idea/dataSources.xml  and add back .idea/sqldialects.xml 
Go to the file changed tab to see exactly which files were changed

PaulMicu (approved):

PR #193, opened 2021-01-21
Fix for rows change scene mechanism

### LINKED ISSUE #186: Bug: List view empty cell
**Describe the bug**
In the list view, trying to access an empty cell which does not hold a reference to any asset will direct to the template "System view". In the system logs, an error will be triggered. See attached pictures.

**To Reproduce**
1. Go to 'List'
2. Click on an empty cell
3. See template for System view
4. See error in system logs

**Expected behavior**
Clicking on an empty cell should not redirect to another view. Expected behaviour is no change at all.

**Screenshots**
image (1)
image

rdsaliba:
Empty rows are now non-clickable


PaulMicu (approved):
seems fine nothing really to review here

najimgh:
Good change, I can no longer click the rows that are empty in the systems list.

najimgh (approved):

PR #195, opened 2021-01-21
#48 delete a system type

### LINKED ISSUE #10: LSTM - Create a network model

### LINKED ISSUE #48: (F11) Remove a system type

PaulMicu:
You might need to implement the fix from : https://jeffreyeverhart.com/2017/11/04/mysql-workbench-error-code-2013-lost-connection-mysql-server-query/

This is because there are a lot of measurements and assets to delete in the background when deleting an asset type.
I had to do it for running the sql in the workbench, it might not be necessary when testing the application itself.

When confirming the deletion , it might take a little time to run the delete and go back to the main screen. This is "normal" and we will need to implement a loading screen in the future for all of these kinds of actions that require a load time

This PR resolves #48

### DIFF
--- a/JavaFXUserInterface/src/main/java/SystemTypeController.java
+++ b/JavaFXUserInterface/src/main/java/SystemTypeController.java
@@ -87,7 +87,7 @@ public class SystemTypeController implements Initializable {
             advisory_value = assetTypeDOA.getAssetTypeBoundary(i + 1, 3);
             ok_value = assetTypeDOA.getAssetTypeBoundary(i + 1, 4);
 
-            systemtypelist.add(new SystemTypeList(systemtypelistname.get(i).getName(),systemtypelistcount.get(i),ok_value,advisory_value,caution_value,warning_value,fail_value));
+            systemtypelist.add(new SystemTypeList(systemtypelistname.get(i).getId(),systemtypelistname.get(i).getName(),systemtypelistcount.get(i),ok_value,advisory_value,caution_value,warning_value,fail_value));
         }
 
         return systemtypelist;
--- a/JavaFXUserInterface/src/main/java/SystemTypeInfoController.java
+++ b/JavaFXUserInterface/src/main/java/SystemTypeInfoController.java
@@ -1,13 +1,18 @@
 import javafx.fxml.FXML;
 import javafx.fxml.Initializable;
+import javafx.scene.control.Alert;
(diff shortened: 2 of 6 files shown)


najimgh:
Upon deleting a System Type I get the following error
Screen Shot 2021-01-22 at 10 37 37 PM

najimgh (changes requested):
Upon deleting a system type I get the following error: (Note the latest SQL script that I ran was **V8**)
Screen Shot 2021-01-22 at 10 37 37 PM

PaulMicu:
the latest script is V8.1 try that one

rdsaliba:
Hi Paul, I'm currently running the latest dump file and still another issue got raised from my side. Please see attached screenshot. I was testing from the application.

Screen Shot 2021-01-24 at 1 27 11 PM

PaulMicu:
@rdsaliba you're not running the lastest version of mariaDB, either your running an older version or you're running mysql. Every time you get an error like this it means that your db is not correct

rdsaliba:
Currently running MariaDB 10.5.8 with script V8.1

jeffbuonamici:
I'm running the latest DB V8.1, and it works for me.

rdsaliba (approved):

jeffbuonamici (approved):

PR #199, opened 2021-01-25
89 establish a system type threshold

### LINKED ISSUE #89: Establish a system type threshold boundaries for RUL values
As a user, I want to establish the threshold boundaries for system type recommendations.

Acceptance Criteria:
- [x] User can define the threshold values of each recommendation boundary for a specific asset type (through UI)
- [x] The thresholds are expressed as integer (RUL)
- [x] The threshold values for an asset type are stored in the database
- [ ] Unit tests
- [x] The threshold names/recommendation labels are fixed by the application

Boundary values are defined for a specific system type and can be updated by the user. A table needs to be added to the database for this.

PO precision (16-11-2020):
The RUL value thresholds for different maintenance recommendations such as good, to be repaired, or to be replaced are determined by the user. These values are saved in the database and compared to a specific asset RUL value to establish the displayed recommendation for that asset.

PaulMicu:
Implemented the saving modification and removal of thresholds, modified add system type to incorporate changes and made some minor changes to the UI 

resolves #89


jeffbuonamici (approved):
Good job.

JavaFXUserInterface/src/main/java/SystemTypeController.java
```diff
@@ -102,12 +100,13 @@ public void fillSystemTypeTable() {
...
-        columnCaution.setCellValueFactory(new PropertyValueFactory<>("value_caution"));
-        columnWarning.setCellValueFactory(new PropertyValueFactory<>("value_warning"));
-        columnFailed.setCellValueFactory(new PropertyValueFactory<>("value_failed"));
+        columnLiveAssets.setCellValueFactory(new PropertyValueFactory<>("liveAssets"));
+        columnArchivedAssets.setCellValueFactory(new PropertyValueFactory<>("archivedAssets"));
+        columnOk.setCellValueFactory(new PropertyValueFactory<>("valueOk"));
+        columnAdvisory.setCellValueFactory(new PropertyValueFactory<>("valueAdvisory"));
+        columnCaution.setCellValueFactory(new PropertyValueFactory<>("valueCaution"));
+        columnWarning.setCellValueFactory(new PropertyValueFactory<>("valueWarning"));
+        columnFailed.setCellValueFactory(new PropertyValueFactory<>("valueFailed"));
```
JayChouteau: I feel like it would increase the readability of the code if we move all the settings of valuefactory into a different method and call this function from attach events. It is not "extracting" a method to properly say, but it is more regrouping those similar actions under a method that has a self-decriptive name for that set of actions. AttachEvents should only make calls to other functions if possible.

JavaFXUserInterface/src/main/java/SystemTypeController.java
```diff
@@ -118,4 +117,6 @@ public void attachEvents() {
         //Attach link to addTypeBtn to go to AddSystemType.fxml
         addTypeBtn.setOnMouseClicked(mouseEvent -> uiUtilities.changeScene(mouseEvent, "/AddSystemType"));
     }
+
+
```
JayChouteau: Minor: those 2 lines can be removed.

JavaFXUserInterface/src/main/java/SystemTypeController.java
```diff
@@ -54,42 +54,40 @@ public void fillSystemTypeTable() {
...
     /**
      * This method will return an ObservableList of SystemList objects
      *
      * @author Shirwa
+     *
+     * edit: There was an issue where the getAssetTypeIdCount() would not match in size to the assetTypeList()
+     * and it would crash the system if there was an asset type with no assets associated to it
+     * so this methode was rewrote
+     *
+     * @autor Paul
```
JayChouteau: Wrong tag -> author

JavaFXUserInterface/src/main/java/AddSystemTypeController.java
```diff
@@ -2,20 +2,18 @@
...
-import javafx.scene.shape.StrokeType;
-import javafx.scene.text.Font;
-import javafx.scene.text.Text;
 import local.AssetTypeDAOImpl;
 
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.ResourceBundle;
 
+
```
JayChouteau: Minor: remove whitespace

JavaFXUserInterface/src/main/java/SystemTypeList.java
```diff
@@ -1,95 +1,152 @@
...
+
+import java.util.ArrayList;
 
 public class SystemTypeList {
-    private SimpleStringProperty name, id;
-    private int associated_assets;
-    private double value_ok,value_caution,value_advisory,value_warning,value_failed;
+    private AssetType assetType;
+    private int liveAssets, archivedAssets;
+    private String valueOk, valueCaution, valueAdvisory, valueWarning, valueFailed , id ;
```
JayChouteau: Is the ID needed? It isn't used anywhere and not included in copy constructor either.

JavaFXUserInterface/src/main/java/SystemTypeInfoController.java
```diff
@@ -60,32 +62,111 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
...
+            assetTypeDAO.updateAssetType(assetType.toAssetType());
+            uiUtilities.changeScene(mouseEvent, "/SystemTypeList");
+        });
+
+        systemTypeName.textProperty().addListener((obs, oldText, newText) -> {
+            if( handleTextChange(newText,originalAssetType.getName()))
+                assetType.getAssetType().setName(newText);
+        });
+        systemTypeDesc.textProperty().addListener((obs, oldText, newText) -> {
+            if( handleTextChange(newText,originalAssetType.getDescription()))
```
JayChouteau: When calling the handleTextChange for description, it seems like the application is trying to parse a double. Maybe there is something wrong going in the if conditions of the method.
SetDescription issue

JavaFXUserInterface/src/main/java/SystemTypeInfoController.java
```diff
@@ -60,32 +62,111 @@ public void initialize(URL url, ResourceBundle resourceBundle) {
...
+
+    }
+
+    /**
+     *  Handle the text change of the user fields to turn on or off the save functionality
+     *
+     * @author  Paul
+     */
+    private boolean handleTextChange(String newText, String field) {
+        if (field.equals(originalAssetType.getName()) || field.equals(originalAssetType.getDescription()) && newText.equals(field))
```
JayChouteau: Behaviour: when editing the description or the name, the save button stays disabled. Editing description is throwing the double.parse error while editing the name doesn't throw an exception, but doesn't enable the save button.
Changing name_can't save
SetDescription_no save

JayChouteau (changes requested):
This is a good chunk of code that is added/updated. Overall, very good job on the logic, the update of the UI handling, the addition of UI utilities, and the way we can edit a system type. There are a few refactoring well done when it comes to DAOs and UIs. The main purpose of this pr, which is to establish the thresholds works great! The user can input his thresholds only as of the form of integer value and those values are properly saved in the database. I have some minor (format) review comments as well as some bigger behavioral review comments. The biggest are the following:

1) The save info button stays disabled when editing the system type name or description. Furthermore, it is throwing an exception when editing the description (see snippets).

2) This wasn't part of the acceptance criteria, but I feel that having some relative boundaries on the threshold values would be important. It might be something we overlooked when defining the issue and acceptance criteria, but it doesn't make sense to have an "OK" threshold that is lower than the "Warning" threshold. That's why I wrote relative boundaries because the threshold should be bound by the other threshold values. That could be implemented either when trying to save or when typing in the value.

JavaFXUserInterface/src/main/java/SystemTypeController.java
PaulMicu: if it helps, sure

JayChouteau:
After discussion with @PaulMicu (assignee on this PR), it was decided that the enforcement of relative boundaries was out of scope for this pr and will be added in a future issue.

JayChouteau (approved):
All the comments from the first review were addressed. Very good job.

PR #200, opened 2021-01-26
#47 maintenance recommendation

### LINKED ISSUE #47: (F5) Suggest Maintenance Recommendation
As a user, I want to see a selected system maintenance information so that I know if an action is required.

Acceptance Criteria:
All the following displays are corresponding to the chosen trained model for that asset type
- [x] Establish (calculate) the recommendation to be shown based on the RUL thresholds given for that system type and the RUL value calculated for that system
- [x] Having a display of the action to perform on the asset (UI) - Asset-grid view
- [x] Having a display of the action to perform on the asset (UI) - Asset-list view
- [x] Having a display of the action to perform on the asset (UI) - Asset information view
- [x] Update Mock-ups
- [x] every time an RUL is calculated, the recommendation should be updated
- [x] Unit tests

PO precision:
The RUL value thresholds for different maintenance recommendations such as good, to be repaired, or to be replaced are determined by the user. These values are saved in the database and compared to a specific asset RUL value to establish the displayed recommendation for that asset. 

This issue needs to compare the RUL value to the threshold, establish the system maintenance recommendation, and display it.

jeffbuonamici:
**New dump file required: cbms_scriptsV10**

Calculates a recommendation based on the Asset Type thresholds. Then displays that recommendation when viewing a system.


JavaFXUserInterface/src/main/java/SystemsController.java
```diff
@@ -27,6 +27,7 @@
 import javafx.util.Duration;
 import local.AssetTypeDAOImpl;
 import local.ModelDAOImpl;
+import org.nd4j.shade.jackson.databind.annotation.JsonAppend;
```
JayChouteau: Unused import to remove.

Server/src/main/java/local/AssetTypeDAOImpl.java
```diff
@@ -58,6 +61,20 @@ public String getAssetTypeBoundary(String asset_type_id, String boundary_type){
         return boundary;
     }
 
+    public HashMap getAssetTypeBoundaries(String asset_type_id) {
```
JayChouteau: The return type of this function is a raw use of hashMaps. It would be better to parameterized the generic type.

Server/src/main/java/local/AssetTypeDAOImpl.java
```diff
@@ -5,13 +5,16 @@
...
+import java.util.HashMap;
 
 public class AssetTypeDAOImpl extends DAO implements AssetTypeDAO {
     private static final String INSERT_ASSET_TYPE = "INSERT INTO asset_type (name, description) values( ?,? )";
     private static final String INSERT_ASSET_TYPE_PARAMETERS = "INSERT INTO asset_type_parameters (asset_type_id, parameter_name, boundary) values(?, ?, ?)";
     private static final String GET_ASSET_TYPES = "SELECT * FROM asset_type";
     private static final String GET_ASSET_TYPE_NAME_FROM_ID = "SELECT name FROM asset_type where asset_type_id = ?";
     private static final String GET_ASSET_TYPE_BOUNDARY = "SELECT *  FROM asset_type_parameters WHERE parameter_name = ? AND asset_type_id = ?";
+    private static final String GET_ASSET_TYPE_BOUNDARIES = "SELECT *  FROM asset_type_parameters WHERE asset_type_id = ?";
+    private static final String GET_ASSET_TYPE_ID_COUNT = "SELECT asset_type_id, COUNT(*) as 'count' FROM asset WHERE archived = 0 GROUP BY asset_type_id";
```
JayChouteau, line 17: This makes 2 String with the same name. Are they doing the same thing? From the develop branch, the statement on line 19 is present in the develop branch while the one on line 17 is added in this pr

Server/src/main/java/local/AssetTypeDAOImpl.java
```diff
@@ -58,6 +61,20 @@ public String getAssetTypeBoundary(String asset_type_id, String boundary_type){
...
+            ps.setString(1, asset_type_id);
+            rs = ps.executeQuery();
+            while (rs.next())
+                boundaries.put(rs.getString("parameter_name"), rs.getDouble("boundary"));
+        } catch (SQLException e) {
+            e.printStackTrace();
+        }
+        return boundaries;
+    }
+
```
JayChouteau, line 83: This comment is more of a discussion, but wouldn't it be more performant to use an ArrayList for this kind of usage. The retrieval of information can differ in time complexity for the worst-case scenario O(1) vs O(n). Hash maps tend to take more memory space as well.  In the case of an ArrayList, it would be a list of boundary object that hold 2 attributes, a boundary name and value.

Server/src/main/java/rul/assessment/AssessmentController.java
```diff
@@ -7,15 +7,18 @@
  */
 package rul.assessment;
 
+import app.item.Asset;
```
JayChouteau: Unused import to remove

Server/src/main/java/rul/assessment/AssessmentController.java
```diff
@@ -27,5 +30,9 @@ public double estimateRUL(Instances testData, Classifier classifier) throws Exce
         return healthAssessment.predictRUL(testData, classifier);
     }
 
+    public String getRecommendation(Double rulEstimation, String assetTypeID) {
+        return recommendationAssessment.getRecommendation(rulEstimation, assetTypeID);
+    }
+
```
JayChouteau: Extra whitespaces to be removed.

Server/src/main/java/rul/assessment/RecommendationAssessment.java
```diff
@@ -0,0 +1,36 @@
...
+    public RecommendationAssessment() {
+        assetTypeDAO = new AssetTypeDAOImpl();
+    }
+
+    public String getRecommendation(Double rulEstimation, String assetTypeID) {
+        HashMap<String, Double> boundaries = getAssetTypeBoundaries(assetTypeID);
+        return calculateBoundaries(rulEstimation, boundaries);
+    }
+
+    private HashMap getAssetTypeBoundaries(String assetTypeID) {
```
JayChouteau: Raw use of parameterized class HashMaps

JayChouteau: This can cause an issue especially since the variable "boundaries" on line 16 is parameterized. Which means there is an unchecked assignment.

Server/src/main/java/rul/assessment/RecommendationAssessment.java
```diff
@@ -0,0 +1,36 @@
+package rul.assessment;
+
+
+import local.AssetTypeDAOImpl;
+import java.util.HashMap;
+
+public class RecommendationAssessment {
+
+    private AssetTypeDAOImpl assetTypeDAO;
```
JayChouteau: This attribute can be set up as final.

Server/src/main/java/local/AssetDAOImpl.java
```diff
@@ -232,6 +232,17 @@ public void insertAsset(Asset asset) {
         }
     }
 
+    public void updateRecommendation(int assetID, String recommendation) {
```
JayChouteau, line 242: Add the @override tag so this method is shown in the interface that this class extends.

Server/src/main/java/local/AssetDAOImpl.java
```diff
@@ -232,6 +232,17 @@ public void insertAsset(Asset asset) {
...
+        try {
+            PreparedStatement ps = getConnection().prepareStatement(UPDATE_RECOMMENDATION);
```
JayChouteau: SonarQube comment. It is better to surround the prepared statements with a try-with-resources block. This is hinted at by IntelliJ as well. It makes sure that the resources are properly closed after usage. See other methods in the class.

Server/src/main/java/local/AssetTypeDAOImpl.java
JayChouteau: Add the @override tag so that it shows in the interface that this class extends.

JayChouteau (changes requested):
Good work. The recommendations are displaying correctly in both the dashboard and list view of the systems.

There are some minor changes that I would like to see regarding the raw use of parameterized classes, unused imports, or white spaces. There are one major issue and 2 discussion points:

1) Major issue - Compiling error due to the doubling of a String initialization in AssetTypeDAOImpl.

2) Discussion: Hashmap vs Arraylist. See comments in the review, but I feel like in our case, the use of Arraylist would be more performant. What are your thoughts?

3) This may be out of scope for this issue since it seems we didn't add it to the acceptance criteria, but we talked about having color coding for the recommendation. Is this something we would like to include in the acceptance criteria of this issue or push them to a next issue?

jeffbuonamici:
I believe I addressed all your concerns. Thanks for the feedback! Good eye.

JavaFXUserInterface/src/main/java/SystemsController.java
JayChouteau: Resolved with last commit.

Server/src/main/java/local/AssetTypeDAOImpl.java
JayChouteau: Resolved in the last commit.

Server/src/main/java/local/AssetTypeDAOImpl.java
JayChouteau: Resolved in the last commit.

Server/src/main/java/rul/assessment/AssessmentController.java
JayChouteau: Resolved in the last commit.

Server/src/main/java/rul/assessment/RecommendationAssessment.java
JayChouteau: Resolved in the last commit.

Server/src/main/java/local/AssetDAOImpl.java
JayChouteau: Resolved in the last commit.

Server/src/main/java/local/AssetDAOImpl.java
JayChouteau: Resolved in the last commit.

Server/src/main/java/local/AssetTypeDAOImpl.java
JayChouteau: Resolved in the last commit,

Server/src/main/java/rul/assessment/RecommendationAssessment.java
JayChouteau: REsolved in the last commit

Server/src/main/java/rul/assessment/AssessmentController.java
JayChouteau: Resolved in the last commit.

JayChouteau:
Thanks @jeffbuonamici for addressing all my review comments. Since the color-coding of the recommendation labels was not part of the acceptance criteria, it will be taken care of in sprint 7. Also, due to time constraints as we are nearing the end of release 2 and need to deliver a demo to the client as well as documentation, the HashMaps vs ArrayList discussion is pushed to the next sprint. It was decided so since the hash maps are allowing the behavior to work correctly and the change would not present any major challenges in the future. Pull request approved

JayChouteau (approved):

PR #211, opened 2021-01-29
88 real time raw data

### LINKED ISSUE #88: Real-Time: Display Assets raw data

jeffbuonamici:
Real time raw data of the sensors now shows in the SystemInfo view.

This resolves #88.

### DIFF
--- a/JavaFXUserInterface/src/main/java/SystemInfoController.java
+++ b/JavaFXUserInterface/src/main/java/SystemInfoController.java
@@ -86,9 +86,8 @@ public class SystemInfoController implements Initializable {
      * Initialize runs before the scene is displayed.
      * It initializes elements and data in the scene.
      *
-     * @param url url to be used
+     * @param url            url to be used
      * @param resourceBundle url to be used
-     *
      * @author Jeff
      */
     @Override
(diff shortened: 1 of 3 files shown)


JayChouteau (approved):
The raw data does indeed get updated and displayed in real-time when we run the real-time demo module alongside the server and UI. Nice job on the timeline addition. Performance is still slow, but that is not specific to that PR.

Also, out of pr, but thanks for integrating the small fix on the get asset type count query. Not much to say otherwise, approved.

PR #245, opened 2021-02-07
Model #3 - Random Forest

Khal-Shah:
This Pull Request includes the implementation of model Random Forest (#181), including the unit test.

To be added to the Database: model_id = 3 and name = RandomForest

Small refactoring: LSTMRegressionTest changed to just LSTMTest as there's no such thing as LSTMRegression.

PR #246, opened 2021-02-07
Model #4 - Random Committee

### LINKED ISSUE #223: RUL model "Random Committee" integration
As a user, I would like to be able to have multiple RUL models to estimate an asset's RUL, to choose from and compare to other models.

Acceptance Criteria:
- [x]  Implement new model strategy
- [x]  Needed dependencies and plugins are added to the project
- [ ]  Documentation on the model variables that can be modified by the user
- [x]  Unit tests

The RUL model implemented (Random Committee) is the one that is ranked 2nd after the research on RUL models (#159).

Khal-Shah:
This Pull Request includes the implementation of model Random Committee (#223), including the unit test.

To be added to the Database: model_id = 4 and name = RandomCommittee

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ cbms.iml
 
 .idea/encodings.xml
 .idea/jarRepositories.xml
+.idea/sqldialects.xml
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,2 +1,3 @@
 # Default ignored files
 /workspace.xml
+/sqldialects.xml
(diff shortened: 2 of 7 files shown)


PaulMicu (approved):
With the exception of the feedback @JayChouteau already gave in the other PR (#245 ), this looks good to me, i think it's would be a better idea to include a small sql script to add the new model in the DB in this PR but i'm also ok with doing it a little latter in a bigger dump. I will still approve the PR as is tho

PR #254, opened 2021-02-08
#247 model too large for db storage

### LINKED ISSUE #247: BUG - LSTM trained models too big to be saved in the database
**Describe the bug**
The models that are trained with the LSTM model cannot be saved into the database. The size of those trained models is too big for it to be saved in the _trained_model_ table as a _serialized_model_ BLOB object.

**To Reproduce**
1. Make sure the models are set to retrain '1' in the database
2. Make sure the application is using the LSTM model as the default model
3. Run the application
4. The error should happen then

**Expected behavior**
When using LSTM to train the models, the models should be saved in the database as a serialized object in the trained model table. That way, the trained models can be fetched and used to calculate an asset RUL.

**Screenshots**
image

 - OS: both Windows and Mac

PaulMicu:
2 issues were found during the bug investigation
1 - after some unknown changes, the train model no longer works as intended
2 - the Classifier object created by the LSTM model is larger than 64k which is what the BLOB column can store

Issue 1 was fixed with rewriting the createAssetInfo() function in the AssetDAOImpl
Issue 2 was fixed in a new dump file by changing the Column type from BLOB to LONGBLOLB for the "serialized_model" column. This brings the available space from 64kb to 4gb  

this issue resolves #247

### DIFF
--- a/Server/src/main/java/app/ModelController.java
+++ b/Server/src/main/java/app/ModelController.java
@@ -59,6 +59,9 @@ public class ModelController {
                 System.out.println("ModelController - initialize - checkAsset - start");
                 checkAssets();
                 System.out.println("ModelController - initialize - checkAsset - end");
+                System.out.println("ModelController - initialize - checkModels - start");
+                checkModels();
+                System.out.println("ModelController - initialize - checkModels - end");
             }
         }, 0, 5000);
     }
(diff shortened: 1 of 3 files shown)


Khal-Shah:
Had no problem running the app using the new dump file. This PR seems to have all the changes needed to fix the database limitation issue for LSTM. The current epoch parameter of LSTM was only set to 1 (one forward/backward passthrough the entire dataset), meaning the size required to hold the model will only increase from the current implementation, so increasing the capacity to 4gb from 64kb was very much needed. 

Approved.

Khal-Shah (approved):

PR #255, opened 2021-02-08
#234 performance enhancements asset information page

### LINKED ISSUE #234: Performance Enhancements - Asset Information page
As a user, I would like to use the application fluidly.

Description:
This issue aims at increasing the application performance through different enhancements. These enhancements are targetting the asset information/detail page. The different issues raised below were hinted at by us by both the performance testing done through release 2 (see release 2 document) as well as by team discussion and evaluation of the situation.

Issues:
- Raw Data List is not taking advantage of JavaFX. generateRawDataTable() in SystemInfoController.java calls createAssetInfo(int assetID) from AssetDaoImpl and queries database every 3 seconds which makes too many calls to the database
- Sensor graph is wrong when there are less than 5 measurements

Acceptance Criteria:
- [x] Leverage the observable list from JavaFX to improve the Raw Data List.
- [x] Reduce the number of calls made by the Graph generations (contructSensorPanes()).
- [x] Graphs are shown correctly for sensors when they have less than 5 measurements. (If there are less than 5 measurements, we need to display them in the correct order).

PaulMicu:
Adjusted how the raw data and sensor pane generation is done in an effort to improve performance and optimize timeline usage
this pull request resolves #234


Khal-Shah (approved):
The refactoring of generateRawDataTable along with the updateRawTableView implementation definitely has a noticeable effect on the asset info page, it feels a lot smoother and more responsive now.  

Approved. 

Edit: has some conflicts after the recent merges of the other branches. Should be good after these are resolved.

JayChouteau (approved):
This pull request does make more use of the observable property of JavaFx, doing fewer calls to the database. The implementation is more robust and more performant. Good job on the effort. 

There are still some merging conflicts to resolve

PaulMicu:
Fixed merge conflicts

JavaFXUserInterface/src/main/java/Controllers/SystemInfoController.java
```diff
@@ -188,15 +201,19 @@ else if (series.getData().size() != measurements.size()) {
...
      */
     public void attachEvents() {
-        systemMenuBtn.setOnMouseClicked(mouseEvent -> uiUtilities.changeScene(mouseEvent, "/Systems"));
-        //Attach link to systemTypeMenuBtn to go to Utilities.SystemTypeList.fxml
+        systemMenuBtn.setOnMouseClicked(mouseEvent -> {
+            timelines.forEach(Timeline::stop);
+            uiUtilities.changeScene(mouseEvent, "/Systems");
+        });
+        //Attach link to systemTypeMenuBtn to go to SystemTypeList.fxml
         systemTypeMenuBtn.setOnMouseClicked(mouseEvent -> uiUtilities.changeScene(mouseEvent, "/SystemTypeList"));
```
jeffbuonamici: Timelines need to be stopped here.

JavaFXUserInterface/src/main/java/Controllers/SystemInfoController.java
```diff
@@ -188,15 +201,19 @@ else if (series.getData().size() != measurements.size()) {
...
         deleteBtn.setOnMouseClicked(this::deleteDialog);
```
jeffbuonamici, line 207: Timelines need to also be stopped here when the page redirects to the Systems page.

jeffbuonamici (changes requested):

JavaFXUserInterface/src/main/java/Controllers/SystemInfoController.java
PaulMicu: this is what  timelines.forEach(Timeline::stop); does

JavaFXUserInterface/src/main/java/Controllers/SystemInfoController.java
PaulMicu: this is what  timelines.forEach(Timeline::stop); does

PaulMicu:
Ran the Jprofiler tool to analize the changes, it doesn't look like there's a major difference in the stats but from actual app usability it's a lot better  
[assetdetail.zip](https://github.com/rdsaliba/notorious-eng/files/5974197/assetdetail.zip)

PaulMicu:
@jeffbuonamici  i addressed your comments

jeffbuonamici (approved):

Issue #238, opened 2021-02-03
General Enhancements - Code Convention Refactoring


JayChouteau:
As a developer, I want to have consistent, clear, and comprehensive code that follows a structure.

Acceptance Criteria:

1) Establish and enforce consistent naming between the documentation, UI and Code. Examples:
- [x] boundaries -> thresholds
- [x] Systems -> Assets
- [x] System Types -> Asset Types
- [x] Sensor -> Attribute

2) Establish and enforce consistent naming and conventions for the use of String constants. Team Consensus was:
- [x] If constant is used in multiple files, get all the string constants in the text constants file
- [x] If constant is used in 1 file, get the string constant at the top of the file
- [x] If constant is a display text Constant, put it in the text constants file

3)  Establish and enforce consistent comment convention. Team consensus:
- [x] Comments at the beginning of a file as it seems documentation asks for it (quick summary of the class, remove versions).
- [x] Javadoc comments before each method that explains what it does. (Not stating the obvious, but giving context or needed explanation).

najimgh:
Note: Investigate how to apply code convention for the whole project.
- What needs to be changed to the code convention for the project?

najimgh:
I need all of you to check some functions and add header comments if you are the one who created/modified it.
To do that, you’ll need to check out the following branch: 238_code_convention_refactoring
**Once you have added header comments let me know for which ones you have done so. This will help me keep track of what has been done or not.** 
Here is a list of functions which need header comments:
- [x] generateRawDataTable (AssetInfoController.java)
~- [ ] setCellValue (AssetInfoController.java)~
- [x] toAssetType (AssetTypeList.java)
- [x] getMeasurementsFromID (MeasurementDAO.java)
- [x] deleteMeasurementsFromID (MeasurementDAO.java)
- [x] insertMeasurement (MeasurementDAO.java)
- [x] setAssetUpdate (AssetDAOImpl.java)
- [x] getAssetTypeThreshold (AssetTypeDAOImpl.java)
- [x] insertAssetType (AssetTypeDAOImpl.java)
- [x] getAssetTypeList (AssetTypeDAOImpl.java)
- [x] getNameFromID (AssetTypeDAOImpl.java)
- [x] updateAssetType (AssetTypeDAOImpl.java)
- [x] deleteAssetTypeByID (AssetTypeDAOImpl.java)
- [x] getLastXMeasurementsByAssetIDAndAttributeID (AttributeDAOImpl.java)
[...]

PR #256, opened 2021-02-08
#239 general enhancements other code enhancements small fixes

### LINKED ISSUE #239: General enhancements - Other code enhancements / small fixes
As a developer, I want to have consistent, clear, efficient, and comprehensive code that follows a structure.

To Be Discussed Again:
- [x] SQL Queries should be in the module where they are used.
- [x] Fix how the recommendation calculation is determined in RecommendationAssessment.java. The threshold value set for a label should be the upper value of that threshold. (Make sure the function handles null values).
- [x] Mouse event in generateThumbnails() in SystemsController.java. Create change scene method in UI Utilities.
- [x] Have DecimalFormat used only once instead of multiple times in iniData() in SystemTypeInfoController.java (Use Object in TextConstants)

PaulMicu:
multiple fixes as described in the original issue #239 
 - queries moved 
 - recommendation determination rewrote
 - mouse event adjusted
 - Decimal Format optimized

resolves #239


JavaFXUserInterface/src/main/java/Controllers/SystemTypeInfoController.java
```diff
@@ -3,22 +3,20 @@
...
-import local.AssetTypeDAOImpl;
 
 import java.net.URL;
-import java.text.DecimalFormat;
 import java.util.Optional;
 import java.util.ResourceBundle;
 
 public class SystemTypeInfoController implements Initializable {
-    private final String SYSTEM_TYPE_LIST = "/Utilities.SystemTypeList";
+    private final String SYSTEM_TYPE_LIST = "/SystemTypeList";
```
rdsaliba, line 19: This is the second time it is declared. Would be better to move it in the constants file and call it from there.

JavaFXUserInterface/src/main/java/Controllers/AddSystemTypeController.java
```diff
@@ -4,20 +4,20 @@
...
 import javafx.scene.control.TextFormatter;
-import local.AssetTypeDAOImpl;
 
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.ResourceBundle;
 
 public class AddSystemTypeController implements Initializable {
-    private final String SYSTEM_TYPE_LIST = "/Utilities.SystemTypeList";
+    private final String SYSTEM_TYPE_LIST = "/SystemTypeList";
```
rdsaliba, line 20: First occurence. Should be declared in the constants file

JavaFXUserInterface/src/main/java/Controllers/SystemsController.java
```diff
@@ -107,7 +99,7 @@ public void attachEvents() {
         systemMenuBtn.setOnMouseClicked(mouseEvent -> uiUtilities.changeScene(mouseEvent, "/Systems"));
 
         //Attach link to systemTypeMenuBtn to go to Utilities.SystemTypeList.fxml
-        systemTypeMenuBtn.setOnMouseClicked(mouseEvent -> uiUtilities.changeScene(mouseEvent, "/Utilities.SystemTypeList"));
+        systemTypeMenuBtn.setOnMouseClicked(mouseEvent -> uiUtilities.changeScene(mouseEvent, "/SystemTypeList"));
```
rdsaliba, line 102: As per my previous comment, this should be called from its constant variable "/SystemTypeList"

PaulMicu:
@rdsaliba yes it should be in the constant file but its out of scope of this issue. We have an issue specifically for that and I figured it will be addressed there. The only reason I even touched those lines was because some of the previous refactoring modified it and broke it.

rdsaliba:
Hi @PaulMicu ,
I went over your changes (there was a lot). I think this covers everything we had discussed in the meeting. Thanks for splitting the commits per what it is changing, made it way easier to review. Let's make sure my comments will be addressed in the other issue you mentioned. Could you referenced it here just to keep track?

rdsaliba (approved):

PaulMicu:
> Hi @PaulMicu ,
> I went over your changes (there was a lot). I think this covers everything we had discussed in the meeting. Thanks for splitting the commits per what it is changing, made it way easier to review. Let's make sure my comments will be addressed in the other issue you mentioned. Could you referenced it here just to keep track?

ya, the issue is  #238

JavaFXUserInterface/src/main/java/Utilities/UIUtilities.java
```diff
@@ -92,22 +92,10 @@ public void changeScene(MouseEvent mouseEvent, String fxmlFileName) {
...
-                Stage window = (Stage) ((Node) mouseEvent.getSource()).getScene().getWindow();
-                window.setScene(systemInfo);
-                SystemInfoController controller = loader.getController();
-                controller.initData(asset);
-                window.show();
-            }
-        } catch (IOException e) {
-            e.printStackTrace();
+        if (!row.isEmpty()) {
+            changeScene(mouseEvent, fxmlFileName, asset);
```
Shirwarsame, line 96: Good idea creating a method for screen changes

Shirwarsame (approved):
@PaulMicu Excellent work. The code works well. As Roy mentioned it was a smart idea to split each commit to its specific issue.

PR #259, opened 2021-02-08
#220 [Refactoring] Testing files organization

### LINKED ISSUE #220: [Refactoring] Testing files organization
As a developer, I would like to have clear and organized testing directories and files.

Description:
The current test files are including both e2e testing and unit testing in different modules. Unit testing and system testing should be in 2 different directories with proper names:

Acceptance criteria:
- [ ] All system tests are in an e2e test folder that is located outside of all the different modules
- [ ] All unit tests are in a test folder that belongs to a module
- [ ] All unit tests located in a module are correspondingly testing that module
- [ ] All imports statements are clear and clean
- [ ] All Unit files have the proper Junit structure (setup, teardown, etc)
- [ ] All system files have the proper TestFx structure (start, etc)
- [ ] All tests can compile

TalalBazerbachi:
- Separated test files in test folders into Unit and System Tests
- Separated system test cases based on the loaded page

resolves [#220]


.idea/misc.xml
```diff
@@ -8,7 +8,7 @@
       </list>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_9" project-jdk-name="11" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_9" default="false" project-jdk-name="15" project-jdk-type="JavaSDK">
```
rdsaliba: This file should not have been part of the commit.

rdsaliba (dismissed):
Thanks Talal for doing this. I see at last the unit tests are well organized.

rdsaliba (changes requested):
need to fix merge conflicts and remove the .idea/ file

TalalBazerbachi:
@rdsaliba I resolved all the issues and updated the branch to be consistent with the develop branch.

rdsaliba:
Hi @t98b, there's still some conflicts that require your attention.

JavaFXUserInterface/src/test/java/UnitTests/AddSystemControllerTest.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public void start (Stage stage) throws Exception {
+        FXMLLoader fxmlLoader = new FXMLLoader(AddSystemController.class.getResource("/AddSystem.fxml"));
+        Parent root = fxmlLoader.load();
+        scene = new Scene(root);
+        addSystemController = (AddSystemController) fxmlLoader.getController();
+        stage.setTitle("CBMS");
+        stage.setScene(scene);
+        stage.show();
+    }
```
JayChouteau: This start method belongs to a system test. To do the 2 tests mentioned below, there is no need to start the application, The setup method and the tests are enough. This should be removed. Also, the file name should be corresponding to the src file name.

Server/src/test/java/UnitTests/preprocessing/reductionTest.java
```diff
@@ -1,7 +1,8 @@
-package preprocessing;
+package UnitTests.preprocessing;
```
JayChouteau: This file should be named with a 1-to-1 correspondence to the src file.

JayChouteau (changes requested):
Separating unit test and system into different packages is good. However, I have a few comments to make:
1) I think that unit test files should be named the same as their corresponding source file (with "Test" at the end). Since unit tests are testing single units, it makes sense to have 1-to-1 correspondence between the test files and the src files.
2) I know we had a discussion about this, but thinking back, I think "e2e" makes more sense as a name for the system tests package. The first reason is that our system tests are actually gonna be e2e tests. The second reason is that this seems to be the convention in the industry  and is also what is used in the template provided by the professor
3) I am unsure about the naming for the e2e test files. The reason is that e2e is supposed to test a full feature. Every feature starts from the main screen and then will perform actions on the main screen or other screens. We could separate by scene as you did, but we would need to agree on the separation, like a test should be located in the scene test file where the last action of the test is performed. That sounds a bit confusing to me. Maybe something like division according to Epics would make more sense: "PrognosticAssessement", "Diagnostic Assessment", or something like that. Or even by user story (like shown in the professor template).

I know I am being very picky here and some of those comments are personal opinions and maybe others could comment on this. Comment nb#3 is also maybe something that could be completed when we work on adding system tests. Feel free to reach out to discuss this.

JayChouteau (approved):

Server/src/test/java/UnitTests/preprocessing/reductionTest.java
JayChouteau: The files were correctly changed to one=to-one correspondence

JavaFXUserInterface/src/test/java/UnitTests/AddSystemControllerTest.java
JayChouteau: This part will be resolved when new unit test and system tests are being implementend

PR #264, opened 2021-02-12
233 performance enhancements

### LINKED ISSUE #233: Performance Enhancements - Assets Page
As a user, I would like to use the application fluidly.

Description:
This issue aims at increasing the application performance through different enhancements. These enhancements are targetting the assets page. The different issues raised below were hinted at by us by both the performance testing done through release 2 (see release 2 document) as well as by team discussion and evaluation of the situation.

Issues:
- getLatestRUL(int assetID) in RulDAOImpl.java is called over a million times
- createAssetInfo(int assetID) in AssetDaoImpl.java is a time hog
- Systems list is not taking advantage of JavaFX utilities (integrated "Observer" Pattern)
- generateThumbnails() in SystemsController.java calls getNameFromId(int assetID) from AssetTypeDaoImpl for each system (Jeff)
- Observable value for RUL in thumbnail (Jeff)

Acceptance Criteria:
- [x] Reduce the number of calls made to getLatestRUL(int assetID) in RulDAOImpl.java to the minimum needed for the application
- [x] Reduce the time hogging of createAssetInfo by reducing the number of calls or what the function calls
- [x] Refactor Systems list logic to take advantage of JavaFX
- [x] Reduce the number of calls made to the database from getNameFromId(int assetID) from AssetTypeDaoImpl for each system.
- [x] Set the RUL value of assets to an observable list to take advantage of JavaFX integrated "Observer" Pattern 
- [x] Use JProfiler for the analysis of the improvements.

jeffbuonamici:
- Reduced time to load application
- Using less threads (1 instead of 2000+)
- Reduced calls to the database
- Takes advantage of observable JavaFX values

This is in regards to issue #233.


JavaFXUserInterface/src/main/java/Controllers/SystemsController.java
```diff
@@ -96,13 +118,22 @@ public void attachEvents() {
...
+            closeTimeline();
+            uiUtilities.changeScene(mouseEvent, "/SystemTypeList");
+        });
 
         //Attach link to addSystemButton to go to AddSystem.fxml
-        addSystemBtn.setOnMouseClicked(mouseEvent -> uiUtilities.changeScene(mouseEvent, "/AddSystem"));
+        addSystemBtn.setOnMouseClicked(mouseEvent -> {
+            closeTimeline();
+            uiUtilities.changeScene(mouseEvent, "/AddSystem");
+        });
```
PaulMicu: In the latest PR i added a function in uiUtilities with the signature: changeScene(ArrayList<Timeline> timelines, MouseEvent mouseEvent, String s) 

it's probably a good idea to use that and pass the timeline to the changescene to keep consistency

PaulMicu (changes requested):
Very good performance improvement, even on my computer.
a few points to addresses:
 - All the changes in AssetDao and AssetDAOImpl should be in the External package in the JavaFXUserInterface Module and not in the Local package of the Server Module
 - The Rul does not appear immediately, in itself not a big deal but if you navigate to the List view before the RULs are displayed in the thumbnail it gives a NullPointerException. So i think we need to change it so it gets the rul as soon as the system page is rendered 

Also, please merge Develop into your branch when opening the PR if the Develop Branch has changed since the branch was opened

JavaFXUserInterface/src/main/java/Controllers/SystemsController.java
jeffbuonamici: I addressed your concerns.

Server/src/main/java/local/AssetDAOImpl.java
```diff
@@ -232,6 +232,31 @@ public void setAssetUpdate(int assetID) {
...
+        newAsset.setLocation(assetsQuery.getString("location"));
+        newAsset.setCategory(assetsQuery.getString("category"));
+        newAsset.setManufacturer(assetsQuery.getString("manufacturer"));
+        newAsset.setSite(assetsQuery.getString("site"));
+        newAsset.setSerialNo(assetsQuery.getString("sn"));
+        newAsset.setRecommendation(assetsQuery.getString("recommendation"));
+        newAsset.setAssetTypeName(assetsQuery.getString("asset_type.name"));
+        //newAsset.setAssetInfo(createAssetInfo(newAsset.getId()));
+        return newAsset;
+    }
```
PaulMicu, line 249: since this is used exclusively in the ui module, it should be in the external package of the ui module not in the server module

PaulMicu (changes requested):

Server/src/main/java/local/AssetDAOImpl.java
jeffbuonamici: createAssetFromQueryResult(ResultSet assetsQuery)

This method is used by 2 other methods in AssetDAOImpl.java in the Server Module.

PaulMicu (approved):

TalalBazerbachi (approved):
The performance of the app is much better now.

PR #269, opened 2021-02-19
252 input validation

### LINKED ISSUE #198: Bug: Add system validation input
**Describe the bug**
There is no validation on the input for the fields

**To Reproduce**
1. Click on 'Add system'
2. Try any combination to fail the add system 

**Expected behavior**

**Screenshots**

### LINKED ISSUE #252: Input Validation
As a user, I want to have validation on all inputs submitted from a form so that it matches the logical expectations.

Description:
Currently, we have Strings and Doubles on the input form. We want to validate the following:

Asset Name:
- Length of 50
- Not Null
- Special Symbols - No Restrictions
- Type Validation - String

Asset Description:
- Length of 300
- Can be Null
- Special Symbols - No Restrictions
- Type Validation - String

Serial Number:
- Length of 20
- Not Null
- Special Symbols - Letters, Numbers, Dash(-)
- Type Validation - String

Manufacturer:
- Length of 20
- Can be Null
- Special Symbols - Letters, Number, Dash(-)
- Type Validation - String

Category:
- Length of 20
- Can be Null
- Special Symbols - Letters, Dash(-)
- Type Validation - String

Site:
- Length of 20
- Can be Null
- Special Symbols - Letters, Numbers, Dash(-)
- Type Validation - String

Location:
- Length of 20
- Can be Null
- Special Symbols - Letters, Numbers, Dash(-)
- Type Validation - String

Asset Type Name:
- Length of 50
- Not Null
- Special Symbols - No Restrictions
- Type Validation - String

Asset Type Description:
- Length of 300
- Can be Null
- Special Symbols - No Restrictions
- Type Validation - String

Thresholds:
- The value of the Thresholds need to follow this logic: Advisory > Caution > Warning > Failure
- Type Validation - Double
- Non-Negative Values
- Can be Null
- Ok does not have a value

Acceptance Criteria:
- [ ] Input field turns red when input is wrong
- [ ] The form does not submit if there is an error
- [ ] Clear explanation to the user as to why the input is wrong

najimgh:
Input validation added for the AddAsset form, AddAssetType form and the AssetTypeInfo form

resolves #252 and #198


TalalBazerbachi (approved):

JavaFXUserInterface/src/main/java/Controllers/AddAssetController.java
```diff
@@ -27,8 +29,8 @@
 
     private static final String SAVE_DIALOG = "Save Dialog";
     private static final String SAVE_HEADER = "Asset has been saved to the database.";
-    private static final String ERROR_DIALOG = "Error Dialog";
-    private static final String ERROR_HEADER = "Please enter values for all text fields.";
+//    private static final String ERROR_DIALOG = "Error Dialog";
+//    private static final String ERROR_HEADER = "Please enter values for all text fields.";
```
JayChouteau: Why is there commented-out code? This should be either deleted or changed.

JavaFXUserInterface/src/main/java/Controllers/AddAssetController.java
```diff
@@ -173,26 +180,103 @@ void saveDialog(MouseEvent mouseEvent) {
...
+//     *
+//     */
+//    void errorDialog() {
+//        Alert alert = new Alert(Alert.AlertType.INFORMATION);
+//        alert.setTitle(ERROR_DIALOG);
+//        alert.setHeaderText(ERROR_HEADER);
+//
+//        alert.showAndWait();
+//    }
+
```
JayChouteau, line 166: Same thing as previous. Commented-out code should be removed or changed.

JavaFXUserInterface/src/main/java/Utilities/UIUtilities.java
```diff
@@ -166,6 +165,115 @@ public void changeScene(MouseEvent mouseEvent, String fxmlFileName, Asset asset)
...
+     * @author Najim
+     */
+    public static void removeInputError(AnchorPane inputError, Text[] errorMessages, boolean[] validInput, TextArea field, int i) {
+        if (errorMessages[i] != null && validInput[i]) {
+            field.getStyleClass().remove("input-error");
+            inputError.getChildren().remove(errorMessages[i]);
+            errorMessages[i] = null;
+        }
+    }
+
```
JayChouteau, line 243: There is a lot of duplication due to the repetition of removeInputError() and createInputError(). The only difference is if the passing argument is a textArea or a textField. It would be better to remove the duplication by having 1 method for each action that allows the passing of an element (either textarea or textfield) and then have a branch condition depending on which one it is within the method. I believe both textarea and textfield are children of TextComponent. This means you could have a TextComponent as a parameter, allowing the function to take in either a textArea or textField and making the distinction within the function.

JavaFXUserInterface/src/main/resources/css/addAsset.css
```diff
@@ -4,4 +4,15 @@
...
+.error-message {
+    -fx-font-weight: bold;
+    -fx-fill: red;
+    -fx-font-size: 11.0;
+}
+
+.input-error {
+    -fx-border-color: red;
+    -fx-border-width: 2px;
 }
```
JayChouteau: There is a lot of duplication between the CSS files for the same code. it would be better to put this on the same CSS file and have it named something more general. Maybe you can reference multiple CSS files in a FXML as well. There is too much duplication.

JavaFXUserInterface/src/main/java/Controllers/AddAssetTypeController.java
```diff
@@ -135,4 +146,89 @@ public boolean saveAssetType(AssetType assetType) {
...
+                UIUtilities.createInputError(inputError, errorMessages, thresholdWarningValue, TextConstants.WARNING_FAILED, 305.0, 0, 5);
+                UIUtilities.createInputError(inputError, errorMessages, thresholdFailedValue, "", 0, 0, 6);
+            } else {
+                validInput[5] = true;
+                validInput[6] = true;
+                UIUtilities.removeInputError(inputError, errorMessages, validInput, thresholdWarningValue, 5);
+                UIUtilities.removeInputError(inputError, errorMessages, validInput, thresholdFailedValue, 6);
+            }
+        }
+        return validForm;
```
JayChouteau, line 256: There is no checking for all the above previous thresholds compared to a threshold, it only compares to the first previous. This means that there are no validation if one value is missing. For example, if your warning threshold has no value, then the failed value can be set higher or lower than the caution value or the ok since the validation on failed is made only compared to warning. It would be important to do the validation compared to the first previous non-empty valid input and not just the first previous.

JavaFXUserInterface/src/main/java/Controllers/AddAssetController.java
```diff
@@ -173,26 +180,103 @@ void saveDialog(MouseEvent mouseEvent) {
...
+        String locationValue = locationInput.getText();
+        String regexWordAndHyphen = "(?=\\S*[-])([a-zA-Z0-9-]*)|([a-zA-Z0-9]*)"; //Any word containing letters, numbers and hyphens
+        String regexLettersAndHyphen = "(?=\\S*[-])([a-zA-Z-]*)|([a-zA-Z]*)"; //Any word containing letters and hyphens
+        double horizontalPosition = 0;
+        boolean validForm = true;
+
+        if (assetNameValue.trim().isEmpty() || assetNameValue.length() > 50) {
+            validForm = false;
+            validInput[0] = false;
+            UIUtilities.createInputError(inputError, errorMessages, assetNameInput, TextConstants.EMPTY_FIELD_ERROR + " and/or \n" + TextConstants.MAX_50_CHARACTERS_ERROR, 65.0, horizontalPosition, 0);
```
JayChouteau: This error message is confusing. It would be better to have those 2 conditions separated and have the corresponding message printed depending on the condition triggered.

JavaFXUserInterface/src/main/java/Utilities/TextConstants.java
```diff
@@ -20,6 +20,17 @@
...
+    //Input Validation Constants
+    public static final String EMPTY_FIELD_ERROR = "Please enter a value";
+    public static final String MAX_20_CHARACTERS_ERROR = "Number of characters has to not exceed 20";
+    public static final String MAX_50_CHARACTERS_ERROR = "Number of characters has to not exceed 50";
+    public static final String MAX_300_CHARACTERS_ERROR = "Number of characters has to not exceed 300";
+    public static final String WORD_HYPHEN_ERROR = "Value can only contain letters, numbers or hyphens";
+    public static final String LETTER_NUMBER_ERROR = "Value can only contain letters or numbers";
+    public static final String ADIVSORY_CAUTION = "The Advisory Threshold needs to be larger \nthan the Caution Threshold";
+    public static final String CAUTION_WARNING = "The Caution Threshold needs to be larger \nthan the Warning Threshold";
+    public static final String WARNING_FAILED = "The Warning Threshold needs to be larger \nthan the Failed Threshold";
```
JayChouteau, line 51: This relates to my previous comment about doing validation on just the previous threshold instead of doing validation on the previous valid threshold. Most likely, the message would be changed.

JayChouteau (changes requested):
The way the input validation errors are highlighted to the user is well done: having the error text fields highlighted in red is clear. However, there are some error messages that are a bit confusing (see comments in the review). All input validations as described in the issue have been performed and work as intended. However, there is one small issue with threshold input validation (see review comments). My last comment is about code duplication. There is a lot of code duplication in this pull-request and it would be better to find ways to get rid of it.

JavaFXUserInterface/src/main/java/Controllers/AddAssetController.java
najimgh: Removed

JavaFXUserInterface/src/main/java/Controllers/AddAssetController.java
najimgh: Removed

JavaFXUserInterface/src/main/java/Utilities/UIUtilities.java
najimgh: Resolved. The function takes a TextInputControl instead of a textArea or TextField

JavaFXUserInterface/src/main/resources/css/addAsset.css
najimgh: Moved CSS styles to the style.css file

JavaFXUserInterface/src/main/java/Controllers/AddAssetController.java
najimgh: Resolved

JavaFXUserInterface/src/main/java/Utilities/TextConstants.java
najimgh: New error messages added

JavaFXUserInterface/src/main/java/Controllers/AddAssetTypeController.java
najimgh: This has been resolved. I have also refactored the code a little more, removing some duplicated code

JavaFXUserInterface/src/main/java/Utilities/UIUtilities.java
JayChouteau: Nice job

JavaFXUserInterface/src/main/java/Utilities/TextConstants.java
JayChouteau: Well done.

JayChouteau (approved):
Thanks for addressing the comments. This pull-request is approved now.

PR #273, opened 2021-02-22
#253 back functionality

### LINKED ISSUE #253: (E1) UI Enhancements - Back Functionality
As a user, I want to be able to move back to the previous view.

Description: 
The user should be able to go back to a previous view.
The view can keep its previous state or not:
   - (Preferred) Previous State Kept: Implement a stack for the views
   - Previous State not Kept: Back button simply goes back to the previous view

Acceptance Criteria:
- [ ] When clicking the back button, the view is changed to the previous view
- [ ] The back button should be placed next to the title of the view (to the left of the title)
- [ ] Clicking on the exit button leaves the application

Shirwarsame:
Adds back functionality and exit program functionality

resolves #253


JavaFXUserInterface/src/main/java/Controllers/AssetTypeController.java
```diff
@@ -124,7 +129,10 @@ public void fillAssetTypeTable() {
     public void attachEvents() {
         //set up the columns in the table
         attachColumnEvents();
-
+        // Change scenes to Assets.fxml
+        backBtn.setOnMouseClicked(mouseEvent -> uiUtilities.changeScene(mouseEvent, TextConstants.ASSETS_SCENE));
```
JayChouteau: I don't think there should be a back button on the main screens of the side left menu. Asset types screen and assets screen is main screens on their own. A back button for coming back to the assets from the asset types is the same as clicking on the left-side menu.

JayChouteau: Also, you could have reached that screen from a specific asset info screen. However, clicking on the back button will always take you back to the assets screen.

rdsaliba (changes requested):
I think the back button should not appear in the side menu. It would be preferable to have on top of each page.

JayChouteau (changes requested):
For most screens, the back button takes you back to where you want to be. However, this is hard-coded and could lead the user to confusion, but this strategy will work for now. However, there is a specific place where it didn't make sense to have a back button. I left a comment on the line.

rdsaliba:
I kind of agree with Jeremie here.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeController.java
JayChouteau: This issue was addressed

JayChouteau (approved):

JayChouteau:
This can be merge as soon as the merge conflicts are resolved

rdsaliba:
Thanks for addressing the issues Shirwa. This is great.

rdsaliba (approved):

PR #274, opened 2021-02-22
#216 evaluating rul models for specific asset type

### LINKED ISSUE #216: (F10) Evaluating RUL models for specific asset type
As a user, I would like to see an RUL model performance for a specific asset type.

Description:
Given a static file filled with asset measurement data from the asset type that the user is consulting, the cbms should be able to train the different models for that asset type with the data provided and should be able to generate the evaluation performance of each model comparing to real RUL values. This is to calculate the RMSE.

Acceptance criteria:
- [x] The system reads data from archived assets for a specific asset type
- [x] The user can specify which ones are for testing and which ones are for training
- [x] Display the number of maximum assets available
- [x] Validate that the number of specified assets is under the maximum
- [x] The system has a generate models button to evaluate all models
- [x] The system has a generate model button for each individual model
- [x] The system generates the trained models and displays the evaluation of each model (RMSE)

TalalBazerbachi:
NOTE: To test, you first need to run the V14 dump file.

The system reads data from archived assets for a specific asset type

The user can specify which ones are for testing and which ones are for training

Display the number of maximum assets available

Validate that the number of specified assets is under the maximum

The system has a generate models button to evaluate all models

The system has a generate model button for each individual model

The system generates the trained models and displays the evaluation of each model (RMSE)

This resolves [#216]


Server/src/main/resources/script/cmbs_scriptsV14/dump_cbms_V14.sql
```diff
@@ -0,0 +1,8 @@
+create table model_evaluation(
+model_id int(11),
+asset_type_id int(11),
+rmse double,
+primary key(model_id,asset_type_id),
+foreign key(model_id) references model(model_id),
+foreign key(asset_type_id) references asset_type(asset_type_id)
+);
```
JayChouteau: This is not technically a dump file. A dump file is supposed to be the current state of the database, meaning that anybody who doesn't have the database setup can run this dump file and get the whole database up and running. Either create a dump file with your addition or give it a meaningful name

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
```diff
@@ -161,6 +206,144 @@ public void attachEvents() {
...
+     * @param mouseEvent is an event trigger to evlauate models
+     * @author Talal
+     */
+    private void evaluateModels(MouseEvent mouseEvent)  {
+        try {
+            ArrayList<String> models = modelDAO.getListOfModels();
+            trainDataset.setClassIndex(trainDataset.numAttributes() - 1);
+            testDataset.setClassIndex(testDataset.numAttributes() - 1);
+            Instances trainSet = populateDataset(trainSize);
+            Instances testSet = populateDataset(testSize);
```
JayChouteau: This variable is never used

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
```diff
@@ -161,6 +206,144 @@ public void attachEvents() {
...
+                        calculateEvaluation(model,trainDataset,testDataset,1);
+
+                    case "LSTM":
+                        ModelStrategy lstmm = new LSTMModelImpl();
+                        Classifier lstm = lstmm.trainModel(trainDataset);
+                        calculateEvaluation(lstm,trainSet,testDataset,2);
+
+                        //                    -------------------------------------------------------------------------------------------
+                        //                    --------------- this part is commented until the rest of the models are added -------------
+                        //                    -------------------------------------------------------------------------------------------
```
JayChouteau: This part should be un-commented since the new models have been integrated

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
```diff
@@ -161,6 +206,144 @@ public void attachEvents() {
...
+                    trainValue.setText(Integer.toString((int)trainSlider.getValue()));
+                    testSlider.setMax(nbOfAssets - (int)trainSlider.getValue());
+                    trainSize = (int)trainSlider.getValue();
+                }
+            });
+            testSlider.valueProperty().addListener(new ChangeListener<Number>() {
+                @Override
+                public void changed(ObservableValue<? extends Number> observableValue, Number number, Number t1) {
+                    testValue.setText(Integer.toString((int)testSlider.getValue()));
+                    trainSlider.setMax(nbOfAssets - (int)testSlider.getValue());
```
JayChouteau, line 227: Given the structure of our application and the processing of data, the sliders should be independent. In our current database, all assets are already identified as being either for training or for testing. This means that the assets that you use for training and for testing are not the same and should not relate to one another. The differentiation is made through the archived attribute in the asset table in the database.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
```diff
@@ -161,6 +206,144 @@ public void attachEvents() {
...
+                }
+            });
+
+            try{
+                modelEvaluateBtn.setOnMouseClicked(mouseEvent -> {
+                    try {
+                        int from = (int) trainSlider.getValue()+1;
+                        int to = (int) trainSlider.getValue()+1+(int) testSlider.getValue();
+                        trainDataset  = DataPrePreprocessorController.getInstance().addRULCol(modelController.createInstancesFromAssets(assetDAO.getAssetsFromAssetTypeID(Integer.parseInt(assetType.getId())).subList(0, (int) trainSlider.getValue()-1)));
+                        testDataset = DataPrePreprocessorController.getInstance().addRULCol(modelController.createInstancesFromAssets(assetDAO.getAssetsFromAssetTypeID(Integer.parseInt(assetType.getId())).subList(from, to-1)));
```
JayChouteau, line 238: The train dataset and test dataset should include assets from the corresponding tag "archived" in the assets table in the database.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
```diff
@@ -161,6 +206,144 @@ public void attachEvents() {
...
+            testDataset.setClassIndex(testDataset.numAttributes() - 1);
+            Instances trainSet = populateDataset(trainSize);
+            Instances testSet = populateDataset(testSize);
+
+            for(int i=0;i<models.size();i++){
+                switch (models.get(i)) {
+                    case "Linear":
+                        LinearRegressionModelImpl linearRegressionModelImpl = new LinearRegressionModelImpl();
+                        Classifier model =  linearRegressionModelImpl.trainModel(trainDataset);
+                        calculateEvaluation(model,trainDataset,testDataset,1);
```
JayChouteau, line 270: This line of code (274) is duplicated a lot of times and could be extracted out of the switch. This could be done by calling it like this : calculateEvaluation(model_clf, trainDataset, testDataset, models.get(i).getID), or something like that at the end of the switch statement.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
```diff
@@ -161,6 +206,144 @@ public void attachEvents() {
...
+
+                    case "LSTM":
+                        ModelStrategy lstmm = new LSTMModelImpl();
+                        Classifier lstm = lstmm.trainModel(trainDataset);
+                        calculateEvaluation(lstm,trainSet,testDataset,2);
```
JayChouteau: The calling of this calculate evaluation is different than the one on line 274 (trainset vs trainDataset)

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
JayChouteau: You could simply call the function without assigning it if you are not using the results.

JayChouteau (changes requested):
Nicely done on the implementation. I left a few comments about the logic. Feel free to reach out to me if you have any questions.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
TalalBazerbachi: @JayChouteau, in one of our meetings we agreed to use the archived data to evaluate the models (for both training and testing).

Server/src/main/resources/script/cmbs_scriptsV14/dump_cbms_V14.sql
TalalBazerbachi: I will change the name of the file so that, other members will know that it's not a dump file.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
TalalBazerbachi: @JayChouteau, in one of our meetings we agreed to use the archived data to evaluate the models (for both training and testing).

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
TalalBazerbachi: I removed the unused Instances objects

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
TalalBazerbachi: I changed the second parameter to the appropriate instances object.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
TalalBazerbachi: I uncommented the part about the new models.

Khal-Shah: As Jeremie said, these models have been integrated and are in the DB as of script V12, We should also try to be using the returned model object from the model classes we already have rather than creating new ones. It is fine for now but when we include parameters in sprint 8, creating them this way will only have the default parameters and not the ones we will be using in the actual model classes.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
```diff
@@ -161,6 +203,132 @@ public void attachEvents() {
...
+            testDataset.setClassIndex(testDataset.numAttributes() - 1);
+
+
+            for(int i=0;i<models.size();i++){
+                switch (models.get(i)) {
+                    case "Linear":
+                        LinearRegressionModelImpl linearRegressionModelImpl = new LinearRegressionModelImpl();
+                        Classifier model =  linearRegressionModelImpl.trainModel(trainDataset);
+                        calculateEvaluation(model,trainDataset,testDataset,1);
+
```
Khal-Shah, line 271: It would be better to use the classes we already have to return the model object rather than creating new ones. It might be fine for now but when we add parameters in sprint 8, creating them this way will only create the default classifier and will not include the parameters added in the actual model classes.

t98b: I am using the classes that we in the models package

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
JayChouteau: @t98b There is still some duplication of the calculationEvaluation

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
JayChouteau: I think we can tackle that during sprint 8 @Khal-Shah since as of now, we only keep 1 trained model per asset type in the database.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
JayChouteau: Appropriate change

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
TalalBazerbachi: I tried calling calculateEvaluation outside the switch statement but it caused errors. because when you declare a classifier you have to assign it. Otherwise it will keep on return the error 'model is always null'.

JayChouteau (approved):

Khal-Shah:
After running the V14 script and the app, the functionalities in the model tab perform as expected. Although currently the RMSE cannot be seen in the GUI, it will be added on a future PR. The system struggles after clicking on the Evaluate button with extreme lagging, but that is to be expected as all the models are being trained. A loading screen or a prompt message during this part would be helpful in the future. 

Approved.

Khal-Shah (approved):

PR #275, opened 2021-02-25
F12 Associating a model to an Asset Type

### LINKED ISSUE #218: (F12) Associating an asset type with a RUL model
As a user, I would like to be able to associate an asset type with a specific RUL model.

Acceptance criteria:
- [x] On the asset type information view - model tab, the user can choose a model to associate with his asset type
- [x] There is a save button to save the selection. The save button will become available once there is a change on the page. 
- [x] Once the selection is done and saved, all assets of that asset type will display the RUL model label chosen on the following screens:
  - [x] Assets list view
  - [x] Asset information view
- [x] Clicking on another page before saving the selection will clear the selection
- [x] Once the selection is done and saved, the chosen RUL model name will be displayed on the asset type information view
- [x] Once the selection is done and saved, the tab refreshes with the model selection highlighted and updated information
- [x] Each asset type is associated with a specific trained RUL model (classifier object) and that connection will be stored in the database

JayChouteau:
This pull request integrates the feature F10, which allows the user to choose which RUL model to choose for a specific Asset Type. Here are the main points of the PR:
0) SQL scripts
- Make sure the latest dump file has been run (v13)
- Run the model_evaluation script on top of that (in folder v14)
1) Try feature
To try out the feature, go to the model tab in Asset Type info view. From there, you will be able to scroll horizontally through the list of models, see their RMSE value, select the model by clicking on the box as well as saving the association by clicking on save. Saving the feature will update the association in the database as well as set the trained_model to be retrained and the assets associated with that asset type to be updated.
2) Evaluation of models
This pull request also includes the evaluation of single models and not necessarily all of them. A button will be available for each model. Clicking on the button will generate the evaluation and update the corresponding RMSE value/values
3) Input Validation
- Models cannot be evaluated as long as the train slider value or test slider value is 0
- Save button cannot be clicked if no model is selected.
4) Verification of the feature
When saving, the model_id should be properly set in the trained_model table for the corresponding asset type. The retrain tag should also be set to true. Finally, the updated tag of all live assets of that asset type should be set to true.
5) UI
The UI looks relatively bad. This was left on purpose to be done by the UI Refactoring that is currently performed by Jeff and avoid duplicate work.
6) REFACTORING
there are 2 refactoring commits in this pull request. These refactoring types are "extract methods"


TalalBazerbachi (changes requested):
when you evaluate models a new list is added instead of just updating the old one.
image

TalalBazerbachi (changes requested):
Saving a model for an asset type is not working properly. 
I tried selecting "Linear" and saving it then I went to another page.
When I went back to the asset type page, the model wasn't selected.

TalalBazerbachi (changes requested):
Last sprint we have agreed to add unit tests for each issue.
So can you add unit tests for the functions that you created?

JayChouteau:
@t98b Here are the answers to your comments:
1 - for updating the thumbnails properly without creating new thumbnails, I left it as is as we agreed in our discussion. This error will be dealt with in your next pull-request that tackles the evaluation of models/UI relation. 
2 - I pushed a commit that solves the issue of keeping track of the current associated model on the page. The associated model is highlighted when the tab is loaded.
3 - I didn't do any unit tests for my methods. We did agree to add them to every issue, so I will be adding them shortly.

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
```diff
@@ -495,4 +555,158 @@ public boolean formInputValidation() {
...
+     * This function also set the retrain tag to true to indicate that the model chosen for the asset
+     * type needs to be retrain. Finally, it labels all live assets of that asset type as updated.
+     *
+     * @author Jeremie
+     */
+    private void saveSelectedModelAssociation() {
+        modelDAO.updateModelAssociatedWithAssetType(selectedModel.getModelID(), assetType.getId());
+        modelDAO.setModelToTrain(assetType.getId());
+        updateAssetsForSelectedModelAssociation(assetType.getId());
+    }
```
Shirwarsame: The database is not updated when the save button is clicked once another model is selected.

Shirwarsame (changes requested):
General issues brought by Talal seem to have been answered. Missing unit tests and new issue found

JavaFXUserInterface/src/main/java/Controllers/AssetTypeInfoController.java
JayChouteau: @Shirwarsame @t98b The reason why the selection was not saved is because the save button could not be clicked in the first place. The calls attached to the save button were good, but the button could never be clicked because there was another UI element on top of it that would prevent the clicking. The element was a transparent anchor pane. I removed it in my latest commit. Sorry for that. Also, clicking on the save button now takes you back to the asset type list. When you click again on the type you modified, the modifications should appear.

JayChouteau:
@t98b @Shirwarsame I added unit tests for the model java file. I am currently adding unit tests for the rest of my code.

JayChouteau:
@t98b @Shirwarsame I added unit tests for several methods implemented. I also extracted some methods of the assettypeinfocontroller into a separate class that I called ModelPanes. This was done to try to preserve the SOLID principles and avoid a Blob Class. Can you take a look at it and give your comments. Thanks!

Shirwarsame (approved):

Shirwarsame (approved):
Excellent work the unit tests work and the changes that you made addressed the issues that were brought up

TalalBazerbachi (approved):

PR #280, opened 2021-03-06
278 parameter db integration

### LINKED ISSUE #278: Design Parameter Handling
As a developer, I want to handle the architecture of Adding/Retrieving Model parameters.

Description:
Models have multiple parameters and each parameter has a different type. To achieve this we will change how the classifier is stored in the database, instead of storing the actual classifier, we want to store the model itself, the model will be able to use weka's internal getters and setters to handle all parameters manipulation. We will also need custom parameters classes and to keep track of the live and eval models in the db

Acceptance Criteria:
- [x] Implement a pattern to handle different types of parameters
- [x] Adjust the trained_model table to handle live and eval models
- [x] Adjust the server module to transition from using the classifier to using the model object when using the db
- [ ] Unit Tests

PaulMicu:
Added a dump file with the added tables and the necessary code to create the models with the parameters from the database. currently there are no db parameters, those will be added in the future

this resolves #278


Khal-Shah:
Ran the new dump file, the new tables for parameters have been created. This is setting up the backend and DB for using parameters and currently has no (UI) impact on the app. One thing that could maybe be improved/clarified is how we are representing the "default" relationship in the tables. As the parameters themselves should not be classified as default, but rather only their values. The parameters we've picked are already applied to all the models, only their values will be changed/edited (from the default values). 

Approved.

Khal-Shah (approved):

jeffbuonamici:
Is this dump file based on the previous ones? Because I believe we had compounded errors due to people not updating their database to the latest version. If there are issues still, maybe we need to setup a quick work session to sort this out.

PaulMicu:
> Is this dump file based on the previous ones? Because I believe we had compounded errors due to people not updating their database to the latest version. If there are issues still, maybe we need to setup a quick work session to sort this out.

I took V13, added the V14 inserts, removed the dataset tables that came back somehow and set the models to train after the dump is run, does this address the errors you're talking about? if not let me know what else we should fix and we'll take the opportunity and do it here

Server/src/main/java/local/ParameterDAOImpl.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ParameterDAOImpl extends DAO implements ParameterDAO {
+    private static final String GET_ALL_PARAMS = "SELECT * from parameter_model_type_value_assoc pmtvs, model_parameter mp, parameter_value pv where pmtvs.model_id = 1 and pmtvs.asset_type_id=1 and pmtvs.parameter_id=mp.parameter_id and pmtvs.value_id=pv.value_id";
```
JayChouteau: The model_id and asset_type_id are set to 1 in the query. This means that no matter which model and asset type are passed, it will always return the parameters for the first ones.

Server/src/main/java/local/ParameterDAO.java
```diff
@@ -0,0 +1,10 @@
+package local;
+
+import app.item.parameter.Parameter;
+
+import java.util.ArrayList;
+import java.util.Map;
```
JayChouteau: unused import statement

Server/src/main/java/local/ParameterDAOImpl.java
```diff
@@ -0,0 +1,65 @@
+package local;
+
+import app.item.parameter.*;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
```
JayChouteau: The map and hashmap imports are unused imports.

Server/src/main/java/app/item/TrainedModel.java
```diff
@@ -46,6 +53,50 @@ public void setModelClassifier(Classifier modelClassifier) {
...
+    }
+
+    public ArrayList<Parameter> getParameter(String name) {
+        return (ArrayList<Parameter>) parameterList
+                .stream()
+                .filter(parameter -> parameter.getParamName().equals(name))
+                .collect(Collectors.toList());
+    }
+
+    public void setParameterList(ArrayList<Parameter> parameterMap) {
```
JayChouteau: parameterMap name should be changed to parameterArrayList or something

Server/src/main/java/app/item/TrainedModel.java
```diff
@@ -6,12 +6,19 @@
  */
 package app.item;
 
+import app.item.parameter.Parameter;
 import weka.classifiers.Classifier;
 
+import java.util.ArrayList;
+import java.util.Map;
```
JayChouteau: unused import statement

Server/src/main/java/app/item/TrainedModel.java
```diff
@@ -46,6 +53,50 @@ public void setModelClassifier(Classifier modelClassifier) {
...
     }
 
+    public ArrayList<Parameter> getParameterList() {
+        return parameterList;
+    }
+
+    public ArrayList<Parameter> getDefaultParameterList() {
+        return (ArrayList<Parameter>) parameterList
+                .stream()
+                .filter(parameter -> parameter.isDefault())
```
JayChouteau: Intellij says that lambda can be replaced with method reference

Server/src/main/java/app/item/TrainedModel.java
```diff
@@ -46,6 +53,50 @@ public void setModelClassifier(Classifier modelClassifier) {
...
+                .collect(Collectors.toList());
+    }
+
+    public ArrayList<Parameter> getLiveParameterList() {
+        return (ArrayList<Parameter>) parameterList
+                .stream()
+                .filter(parameter -> parameter.isLive())
```
JayChouteau: Intellij says that lambda can be replaced with method reference

JayChouteau (changes requested):
The changes to the database and logic are comprehensive are connect well. This is the good groundwork for the model parameters used in the application. Left a few minor comments, but the logic and design is well thought. Good separation of files and logic. Clear to understand.

jeffbuonamici:
> > Is this dump file based on the previous ones? Because I believe we had compounded errors due to people not updating their database to the latest version. If there are issues still, maybe we need to setup a quick work session to sort this out.
> 
> I took V13, added the V14 inserts, removed the dataset tables that came back somehow and set the models to train after the dump is run, does this address the errors you're talking about? if not let me know what else we should fix and we'll take the opportunity and do it here

I'm still going to approve this pull request because this issue isn't responsible for fixing our database. I think we should create a bug report or new issue to review our database. Other than that, all acceptance criteria have been met once you address @JayChouteau 's concerns.

Server/src/main/java/utilities/Constants.java
```diff
@@ -8,6 +8,9 @@
     public static final String WARNING_THRESHOLD = "Warning";
     public static final String FAILED_THRESHOLD = "Failed";
 
+    public static final int STATUS_LIVE = 1;
+    public static final int STATUS_EVALUATION = 2;
```
JayChouteau, line 12: I am guessing these constants are gonna be used in the next issues on model parameters to make use of these?

Server/src/main/java/app/ModelController.java
```diff
@@ -102,7 +102,7 @@ public TrainedModel getModelForAssetType(List<TrainedModel> trainedModels, Strin
             if (tm.getAssetTypeID() == Integer.parseInt(assetTypeID))
                 return tm;
         }
-        trainedModels.add(modelDAOImpl.getModelsByAssetTypeID(assetTypeID));
+        trainedModels.add(modelDAOImpl.getModelsByAssetTypeID(assetTypeID, 1));
```
JayChouteau: I believe this would be a place to be using the constants defined for the status_id in the Constants file

Server/src/main/java/app/ModelController.java
```diff
@@ -143,6 +143,7 @@ private void trainModel(TrainedModel trainedModel) throws Exception {
         ModelStrategy modelStrategy = getModelStrategy(trainedModel);
         if (modelStrategy != null) {
             ModelsController modelsController = new ModelsController(modelStrategy);
+            trainedModel.setModelStrategy(modelStrategy);
```
JayChouteau, line 156: isn't that redundant since you are already getting the model Strategy of the trained model from line 143? It feels redundant.

Server/src/main/java/local/ModelDAOImpl.java
```diff
@@ -32,7 +32,7 @@
      * @author Paul
      */
     @Override
-    public String getModelNameFromAssetTypeID(String assetTypeID) {
+    public String getModelNameFromModelID(String assetTypeID) {
```
JayChouteau: It would be good to update the parameter name and description of the method at the same time.

Server/src/main/java/local/ModelDAOImpl.java
```diff
@@ -72,20 +72,20 @@ public String getModelNameFromAssetTypeID(String assetTypeID) {
      * Given an arraylist of trained models this function will write the new Classifier object
      * to the corresponding trained model entry after an model training phase
      *
-     * @param tms<TrainedModel> represents a list of trained models
+     * @param tm<TrainedModel> represents a list of trained models
      * @author Paul
      */
     @Override
-    public void setModelsToTrain(ArrayList<TrainedModel> tms) {
+    public void setModelsToTrain(TrainedModel tm) {
```
JayChouteau: It would be good to update the method description and param description to reflect the new changes. Also, the name "setModelsToTrain" could be updated to "setModelToTrain" (singular) to add clarity to the function.

Server/src/main/java/local/ModelDAOImpl.java
```diff
@@ -19,10 +19,10 @@
 import java.util.ArrayList;
 
 public class ModelDAOImpl extends DAO implements ModelDAO {
-    private static final String UPDATE_SERIALIZE_OBJECT = "UPDATE trained_model SET retrain = false, serialized_model  = ? WHERE model_id = ? AND asset_type_id = ?";
+    private static final String UPDATE_SERIALIZE_OBJECT = "UPDATE trained_model SET retrain = false, serialized_model  = ? WHERE model_id = ? AND asset_type_id = ? and status_id = ?";
     private static final String GET_SERIALIZE_OBJECT = "SELECT * FROM trained_model WHERE retrain = true";
-    private static final String GET_MODEL_NAME_FROM_ID = "SELECT name from trained_model, model where trained_model.model_id = model.model_id and asset_type_id = ?";
-    private static final String GET_MODEL_FROM_ASSET_TYPE = "SELECT * FROM trained_model WHERE asset_type_id = ?";
+    private static final String GET_MODEL_NAME_FROM_ID = "SELECT name from trained_model, model where trained_model.model_id = model.model_id and model.model_id = ?";
```
JayChouteau: Maybe the SELECT should be changed to a SELECT DISTINCT because of the change from asset type id to model ID matching. The reason is that this query might return multiple times the same value since different asset type might be using the same model strategy in the trained model table

JayChouteau (changes requested):
Very good job on the logic as the groundwork of the model parameters. The change from an interface to an abstract class for the model strategy gives us more flexibility on the classifiers and incoming getting and setting of classifiers. The status ID addition to a trained model will also allow us to separate evaluation to compare the models from the actual trained classifiers used for the assets. I left a few minor review comments below. 

The last concern I have with this is that the addition of a status ID means that there will be some updates needed for the queries and methods used for the association of model and asset type feature. I don't know if this update was included in an already existing issue or if we needed to create one?

Server/src/main/java/utilities/Constants.java
PaulMicu: yep

Server/src/main/java/local/ModelDAOImpl.java
PaulMicu: no, but the trained  model table is no longer necessary, so i'm going to update the query to simply get the name from the model table

Server/src/main/java/app/ModelController.java
PaulMicu: yea, added the constant after, forgot to check, thanks

Server/src/main/java/app/ModelController.java
PaulMicu: Actually the modelStrategy is in the trainedmodel so we're both wrong, i adjusted the code. The get ModelStrategy is no longer used but i have a feeling we will need it in the future so i'm not gonna remove it right now

JayChouteau (approved):
Thanks for addressing the comments I made. Nice job on updating the checkmodel method in Model Controller to be in line with this pr changes. Also, thanks for adressing the createTrainedModelFromResultSet() method in ModelDAOImpl. I was always confused as to why there was a boolean parameter if it was always set to false and how the serialized trained model was read and assigned. Those changes make it a lot clearer and comprehensive for me. Approved.

Khal-Shah (approved):
All required changes are there. Approved.

PR #281, opened 2021-03-09
235 revamp UI

### LINKED ISSUE #235: (E1) UI Enhancements - UI Clean-up
As a user, I want a cleaner and more user-friendly UI interface.

This issue aims at cleaning the code structure and files for the UI as well as improving the look of the UI.

Acceptance Criteria:
- [x] FXML files clean-up: All Styling should be in CSS and disposition should be in JavaCode
- [x] CSS Files clean up: Proper structure and organization of files. Elements are well identified by type or ID
- [x] The navigation bar needs to be its own component instead of being copy-pasted over and over
- [ ] Template for our components (tabs, etc.) so when we need change -> 1 change for all tabs instead of 1 change per tab (to make it easier to update components throughout the project).

jeffbuonamici:
Redesigned the UI.

This satisfies issue #235


PaulMicu (changes requested):
I didn't look too much in the code but a few usability notes:
- In the select a model section of the asset type, when evaluating a model, the list gets duplicated with the result instead of replaced.
The ordering button on the asset page:
1- it's too small to read content It says "Ascen...", it should be large enough to print the whole string
2- the onclick seems to be on the wrong element,the drop down menu only shows up when i click on the top of the box, it should show up when i click anywhere on the box.
- When resizing the window, the asset thumbnail no longer use all the extra space.
- the exit button should not have a static position, it should be in the very bottom left at all times
- On the asset info page, some of the name of the sensors are too big and go outside the box
- Only some of the assets are displayed, it looks like there's a sublist somewhere that needs to be removed

JavaFXUserInterface/src/main/java/Controllers/AssetsController.java
```diff
@@ -191,45 +178,77 @@ public void generateThumbnails() {
...
+            if(asset.getRecommendation().equals("Ok"))
+                statusPane.getStyleClass().add("ok");
+            else if(asset.getRecommendation().equals("Advisory"))
+                statusPane.getStyleClass().add("advisory");
+            else if(asset.getRecommendation().equals("Caution"))
+                statusPane.getStyleClass().add("caution");
+            else if(asset.getRecommendation().equals("Warning"))
+                statusPane.getStyleClass().add("warning");
+            else if(asset.getRecommendation().equals("Failed"))
+                statusPane.getStyleClass().add("failed");
```
PaulMicu: this should be replaced with the constants

JavaFXUserInterface/src/main/java/Utilities/ModelPanes.java
```diff
@@ -42,7 +42,7 @@ public void highlightAssociatedModel(ObservableList<Pane> modelPanes, int modelI
             if ((i + 1) == modelID) {
                 modelPanes.get(i).setStyle("-fx-border-color: red");
             } else if ((i + 1) != modelID) {
-                modelPanes.get(i).setStyle("-fx-background-color: #e0e0eb");
+                //modelPanes.get(i).setStyle("-fx-background-color: #e0e0eb");
```
PaulMicu: this should probably be removed if it's dead code

PaulMicu (changes requested):

najimgh (dismissed):
I have come across much of the same remarks/notes which Paul mentioned. 
However, here are some additional little remarks:
- When going full screen in the "Models" tab on the asset type, there is an amount of empty space under the model cards. It would be better to either remove that empty space or make it so the model cards wrap underneath and fill that empty space.
- I noticed that whenever I am in Full-Screen and click on something in the application, the window exits the Full-Screen size and changes to its default window size.

Otherwise, the UI is much better and more soothing for the eyes.

JayChouteau:
@najimgh  The white space beneath the models in the model tab was meant for displaying the model parameters and playing with them. This was how we first agreed to do it. I don't know if another direction was taken since then, but the last time I discussed it with Jeff and other team members, that was the case.

@PaulMicu For the list duplication of the thumbnails when the evaluation model is called is an issue inherent from my pull request. After a review by Talal and discussion with him, we found a solution that is gonna be implemented with his upcoming pull request to move evaluation to the server. We are aware of the issue and the fix exists on Talal's branch for that issue.

jeffbuonamici:
> * On the asset info page, some of the name of the sensors are too big and go outside the box <

This doesn't happen for me.
image


> * Only some of the assets are displayed, it looks like there's a sublist somewhere that needs to be removed <

This isn't something that was changed in this branch. But I removed it regardless.

Khal-Shah (commented):
This was a very big revamp that changes the look and feel of the UI for the better. 
One thing to note however, the icons are not visible on the left sidebar. If I hover over them the area where they are suppose to be are highlighted and I can click on them (functional) but they are not visible. Is this not happening for anyone else?

If this is a system specific issue (Macs) then I am fine having it as a bug next sprint.

image

jeffbuonamici:
- Fixed Order dropdown not clicking properly
- Asset thumbnails now take up the full space when resizing window
- Exit button is anchored to the bottom instead of staying in same position
- Remove sublist of assets that was there for some reason
- Changed how we change scenes so that the window doesn't resize every time we change pages.

TalalBazerbachi (dismissed):
UI looks much better now!
I don't know if it's part of the issue but the search bar is not working.
also the models page, the height of model infobox should be a bit smaller because it gets cropped.
image

jeffbuonamici:
> UI looks much better now!
> I don't know if it's part of the issue but the search bar is not working.
> also the models page, the height of model infobox should be a bit smaller because it gets cropped.
> image

Search bar is not part of this issue.

Models page looks fine for me:
image

rdsaliba:
Hi @jeffbuonamici,
From my end, I have a serious decrease in performance while launching the app on your branch... It's taking way too much time to load and it bugs each time a click on a tab. The UI looks neat though.

rdsaliba:
> Hi @jeffbuonamici,
> From my end, I have a serious decrease in performance while launching the app on your branch... It's taking way too much time to load and it bugs each time a click on a tab. The UI looks neat though.

You can dismiss this comment. I noticed you removed the sublist on the getAllLiveAssets(0,50). This is why it was that slow for me...

rdsaliba (dismissed):
Hi again @jeffbuonamici 
The new structure and organization of the CSS files is very decoupled which makes a lot of code redundant. I believe a better structure would have been to have the main styling in one file and another one for specifics. Also, all styling should be done through CSS, there's still a lot of in-line styling. I'm assuming this was generated automatically with SceneBuilder, but this should have been removed at this point.

image

JavaFXUserInterface/src/main/resources/AddAssetType.fxml
```diff
@@ -1,150 +1,135 @@
...
+                            <image>
+                                <Image url="@imgs/outline_arrow_back_black_36dp.png"/>
+                            </image>
+                        </ImageView>
+                    </graphic>
+                </Button>
+                <Text fx:id="addAssetTitle" fill="#0c072e" layoutX="62.0" layoutY="51.0" strokeType="OUTSIDE"
+                      strokeWidth="0.0" styleClass="header" text="Add Asset Type">
+                    <font>
+                        <Font name="Segoe UI Bold" size="18.0"/>
```
rdsaliba, line 36: Example of in-line styling.

JavaFXUserInterface/src/main/resources/AddAssetType.fxml
```diff
@@ -1,150 +1,135 @@
...
                 </font>
             </Text>
-            <TextField fx:id="assetTypeName" layoutX="220.0" layoutY="61.0" prefHeight="25.0" prefWidth="350.0" />
-            <Text fill="#0c072e" layoutX="20.0" layoutY="79.0" strokeType="OUTSIDE" strokeWidth="0.0" text="Asset Type Name">
+            <TextField fx:id="assetTypeName" layoutX="233.0" layoutY="67.0" prefHeight="40.0" prefWidth="350.0"/>
+            <Text layoutX="43.0" layoutY="133.0" strokeType="OUTSIDE" strokeWidth="0.0" styleClass="formLabel"
+                  text="Asset Type Description">
                 <font>
-                    <Font name="Segoe UI Bold" size="14.0" />
+                    <Font name="Segoe UI Bold" size="14.0"/>
```
rdsaliba, line 65: Example of In-line styling

JavaFXUserInterface/src/main/resources/css/revamp/assetTypeInfo.css
```diff
@@ -0,0 +1,182 @@
...
+    -fx-background-color: #FFFFFF;
+}
+
+.formLabel {
+    -fx-fill: #3A3A3A;
+    -fx-font-size: 14px;
+    -fx-font-family: 'Segoe UI';
+}
+
+#deleteBtn {
```
JayChouteau, line 11: There is a lot of duplication in the CSS files that could be moved to the same file.

JayChouteau (dismissed):
Overall, very good job on the new design. It definitely looks better and more professional. It is also clearer to use and differentiate elements. Here are some comments about the changes, some are minor fixes:
1. I think it would be good if the "Add asset" and "Add Type" buttons had the exact same disposition (that they could superpose perfectly), so that it doesn't feel like there is a shift when you move from the asset screen to the asset type screen.
2. Label Naming consistency: On the assets screen, the title is 'Systems', but the button says "Add asset". I think we can replace "Systems" with "Assets".
3. The fact that the add buttons and evaluate buttons are the same color as the delete buttons makes it a little confusing. Having different colors for different actions (buttons) makes the UX easier. For example, we could have the evaluate buttons the same colour as the left-side component (blue) and the add button as the same colour as the save button (green). I understand we don't want to have too many colours, that is why I am suggesting those.
4. Very very minor, but I feel like the Evaluate button for all models should have a different label than the other single individual evaluate buttons, like "Evaluate All" or something to make it more explicit.
5. On the Asset Type Information screen, the Title is "Asset info" and should be "Asset Type info"
6. On both the Asset Type information screen and Asset information screen, the labels are not aligned. They should all be either left-side aligned or right-side aligned.
7. Personal Preference, but I would have the order dropdown have "Order By" as a default label.

PaulMicu (approved):

Shirwarsame (approved):
Excellent work Jeff. This revamp is well done and adds a pleasant viewing experience to the user

PR #298, opened 2021-03-21
[BugFix] Null value handling for asset type parameter

JayChouteau:
The initial issue created for this bug was split into 2 different bugs:
1) Deleting an asset type freezes the application
2) Error when trying to display a null value asset type parameter.

This pull-request deals with the second bug (2). Please refer to the issue for all comments on observations and resolution with screenshots.

PR #304, opened 2021-03-23
[Bugfix] Delete action change scene bugfix

### LINKED ISSUE #299: [BUG] Delete Buttons for Assets and Asset types don't lead to the proper screen
**Describe the bug**
This bug has 2 parts, but of the same nature. Deleting an asset or asset type doesn't take the user back to the proper screen after the deletion is completed.

**To Reproduce**
A) Asset Type
1. Go to Asset Type List Screen
2. Click on any asset type
3. Click on Delete Button
4. Click on the Ok button in the dialog box
6. See error - Once the deletion is done, the app brings you back to the deleted asset type info page

B) Asset
1. Go to Asset List Screen
2. Click on any asset
3. Click on Delete Button
4. Click on the Ok button in the dialog box
6. See error - Once the deletion is done, the app brings you back to the deleted asset type info page

**Expected behavior**
Once the delete has been confirmed, the user should be bought back to the the corresponding item list, asset type list for (A) and asset list for (B)

all OS versions

JayChouteau:
This pull request is a bugfix for the issue (#299 ) as well as some quality of life changes and refactoring for the custom dialogs and text constants. Summary of the pull-request:
- Deleting an asset brings the user back to the assets list screen instead of the deleted asset info screen
- Deleting an asset type brings the user back to the asset types list screen instead of the deleted asset type info screen
- When archiving an asset, the time cycles are shown in reverse order, from the latest time cycle to the earliest
- Custom Dialog refactoring - method and variables renaming
- Custom Dialog refactoring - method signature change
- Custom Dialog refactoring - consistency
- Custom Dialog - Switched "Ok" and "Cancel" Buttons to have them consistent with the disposition of such buttons in the application

IMPORTANT: PULL REQUEST #298 needs to be merged before this one


Khal-Shah (approved):
The changes seem functional and working well. Very good work with the refactoring and implementation. As for the UI change, maybe I got used to the old one but I'm not as big of a fan with how the custom dialog box now contrasts with the background of our app. I think a little difference in colors to make it "pop" more wouldn't be a bad idea as this is supposed to be an alert notification after all. 

But that doesn't take anything away from this PR. Approved. 

P.S: potential reviewers make sure to check into BUG-Delete_Asset_**Type_#2** as there is another branch called _BUG-Delete_Asset_Type_ that I mistakenly checked into first due to terminal autocomplete not adding on the _#2.

JayChouteau:
> The changes seem functional and working well. Very good work with the refactoring and implementation. As for the UI change, maybe I got used to the old one but I'm not as big of a fan with how the custom dialog box now contrasts with the background of our app. I think a little difference in colors to make it "pop" more wouldn't be a bad idea as this is supposed to be an alert notification after all.
> 
> But that doesn't take anything away from this PR. Approved.
> 
> P.S: potential reviewers make sure to check into BUG-Delete_Asset_**Type_#2** as there is another branch called _BUG-Delete_Asset_Type_ that I mistakenly checked into first due to terminal autocomplete not adding on the _#2.

I agree with the fact that the custom dialog could be revamped a little bit. That is why there is an issue in the active sprint for that reason. The custom dialog has been in our project for a few sprints now, however.

najimgh (approved):
Agreeing with @Khal-Shah here. The refactorization was done very well, and the changes address the bug issue. I was also going to comment on the custom dialog box and how styling it to fit our application would be better, but I see that it is an issue in the active sprint.

PR #306, opened 2021-03-25
[BugFix + Refactoring] Input validation logic and display

### LINKED ISSUE #301: [BUG] Display errors on input validation
**Describe the bug**
There are some issues with the input validation on all input forms in the app. There are currently 3 places:
- Add Asset
- Add Asset Type
- Asset Type Information
There were 2 main issues for all of these locations:
(A) The error messages for input validation are not aligned properly
(B) Some error messages are not appearing

**To Reproduce**
The below steps are shown for the asset type information page. However, these steps can be reproduced on all locations mentioned above.
(A) Steps to reproduce the behavior:
1. Go to the asset type list screen
2. Click on an asset type from the list and access the asset type information screen
3. Enter a value for caution threshold (that will be lower than the value for the warning threshold)
4. Click on save
5. See error - the input validation messages are not aligned with the proper text fields

(B) Steps to reproduce the behavior:
1. Go to the asset type list screen
2. Click on an asset type from the list and access the asset type information screen
3. Enter a value for the advisory threshold
4. Make sure that the value for the caution threshold is null
5. Make sure that the value for the warning threshold is higher than the value for the caution threshold
6. Click on save
7. See error - You are left on the same asset information page, the save button is enabled, but no error messages appear.

**Expected behavior**
(A) The error messages should be aligned with the text field
(B) An error message should be displayed if a threshold value is not properly ordered compared to the other ones.
OK > Advisory > Caution > Warning > Failed

**Screenshots**
(A) 
(B) 

All Os

**Additional context**
This has to do with the input validation part

JayChouteau:
Solves the bug (#301 ) for the display of the input validation for all input forms in the application. Namely, for the 3 following locations:
- Add asset
- Add asset type
- Asset type information

Furthermore, on top of the bug fixing, some refactoring was done on the logic of the input validation:
- Extracted class: all form input validation and helpers were extracted into a class called FormInputValidation
- Signature change: multiple methods had their signature changed due to the nature of the extraction as well as refactoring on the logic of the inpu validation
- Complexity and duplication reduced: The complexity and duplication were greatly reduced thanks to the 2 previous actions mentioned
- More Consistency and automation: The logic was made more automized and less hard-coded, which will allow for better maintainability. The logic was also made more consistent between the different usages of the input validation.
- Small modification to the calling of the text constants on the code that was touched.
- Comments and description updated.

Please refer to the issue to try out the steps to reproduce the bug for all input validation locations in the app.


TalalBazerbachi (commented):
when displaying one line of input validation, I think the it should be a bit lower.
image

TalalBazerbachi (approved):

JayChouteau:
> when displaying one line of input validation, I think the it should be a bit lower.
> image

I have aligned vertically the error messages and have made the messages more specific to exactly which error is thrown off by the input validation instead of having and/or

verticalAlign_2
verticalAlign

JavaFXUserInterface/src/main/java/utilities/FormInputValidation.java
```diff
@@ -0,0 +1,300 @@
...
+
+    /**
+     * Compares two thresholds and determines if the previous threshold is larger than the next. If the the previous threshold is larger than
+     * the next, the threshold difference is considered invalid and returns false.
+     *
+     * @param previousThreshold The Threshold preceding
+     * @param nextThreshold     The Threshold succeeding
+     * @author Najim
+     */
+    private static boolean compareThresholds(TextField previousThreshold, TextField nextThreshold) {
```
najimgh, line 302: Small warning from Intellij suggesting to invert the method (considered refactoring):

Screen Shot 2021-03-27 at 2 54 21 PM

najimgh (changes requested):
Really nice job on addressing the alignment issue of the error messages. Using a field's original positioning and adding a gap to display the error message made it so there are practically no hard-coded values.
Also, very good job refactoring the input validations into a single utility class. This will provide much more flexibility for future changes. I have left a comment for one further refactoring.

Overall, the bug issue was nicely addressed and you went beyond what was asked by refactoring.

JavaFXUserInterface/src/main/java/utilities/FormInputValidation.java
JayChouteau: I also saw this and thought about it, but the refactoring of that method was really out of scope, and tbh, at that point, I felt like I did enough refactoring for the in put validation. Because , for this specific method, I simply moved it from where it was to this new class, but I didnt touch it. For me, it works fine. I felt like the effort was enough for this pull request

najimgh: Fair enough, it is a small refactorization that doesn't affect much.

najimgh (approved):

PR #307, opened 2021-03-25
221 search bar

### LINKED ISSUE #221: Search Bar - Assets
As a user, I would like to search for specific assets.

Acceptance Criteria:
- [x] Filtering the list of assets based on the input text on the search bar and display the ones that match
- [x] Live Search
- [x] Search for everything (every possible attribute or member of assets)
- [x] Search will be done on top of the applied filters (only applies to the list that has been filtered)
- [x] Search return should keep the current ordering 
- [x] Unit tests
- [x] Code Clean-up

najimgh:
Search Asset Functionality
Assets can be searched by:  

- name 
- rul
- description
- serial number
- manufacturer
- category
- site
- location
- recommendation
- asset type name

This resolves #221

### DIFF
--- a/JavaFXUserInterface/src/main/java/controllers/AssetsController.java
+++ b/JavaFXUserInterface/src/main/java/controllers/AssetsController.java
@@ -258,7 +258,7 @@ public class AssetsController implements Initializable {
      * @return assetsToDisplay
      * @author Najim
      */
-    private ObservableList<Asset> setAssetListToDisplay() {
+    public ObservableList<Asset> setAssetListToDisplay() {
         ObservableList<Asset> assetsToDisplay;
         if (!searchedAssets.isEmpty()) {
             assetsToDisplay = FXCollections.observableArrayList(searchedAssets);
--- a/JavaFXUserInterface/src/main/java/controllers/AssetsController.java
+++ b/JavaFXUserInterface/src/main/java/controllers/AssetsController.java
@@ -225,6 +225,8 @@ public class AssetsController implements Initializable {
         search.textProperty().addListener((observable, oldValue, newValue) -> {
             searchedAssets.clear();
             if (!newValue.equals("")) {
+                //For each asset, we determine if the value from the search is found in the asset's member variables.
+                //Member variables: Rul, name, description, serial number, manufacturer, category, site, location, recommendation and asset type name
                 assets.stream().filter(o -> StringUtils.containsIgnoreCase(o.getRul().get(), newValue) || StringUtils.containsIgnoreCase(o.getName(), newValue)
                         || StringUtils.containsIgnoreCase(o.getDescription(), newValue) || StringUtils.containsIgnoreCase(o.getSerialNo(), newValue)
                         || StringUtils.containsIgnoreCase(o.getManufacturer(), newValue) || StringUtils.containsIgnoreCase(o.getCategory(), newValue)
@@ -244,7 +246,7 @@ public class AssetsController implements Initializable {
             if (assetsTabPane.getSelectionModel().getSelectedItem().getId().equals("thumbnailTab")) {
                 assetsThumbPane.getChildren().clear();
                 generateThumbnails();
-            } else {
+            } else if (assetsTabPane.getSelectionModel().getSelectedItem().getId().equals("listTab")) {
                 assetsListPane.getChildren().clear();
                 generateList();
             }
--- a/JavaFXUserInterface/src/main/java/controllers/AssetsController.java
+++ b/JavaFXUserInterface/src/main/java/controllers/AssetsController.java
@@ -31,6 +31,7 @@ import javafx.scene.layout.HBox;
 import javafx.scene.layout.Pane;
 import javafx.scene.text.Text;
 import javafx.util.Duration;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import rul.assessment.AssessmentController;
@@ -74,8 +75,14 @@ public class AssetsController implements Initializable {
     private Tab listTab;
     @FXML
     private ChoiceBox<String> sortAsset;
+    @FXML
+    private TextField search;
+    @FXML
+    private TabPane assetsTabPane;
+    private String searchMatch = "No Search"; // search states to help display the right asset list: No Search / Match / No Match
     private UIUtilities uiUtilities;
     private ObservableList<Asset> assets;
+    private ObservableList<Asset> searchedAssets;
     private Timeline rulTimeline;
 
     public AssetsController() {
@@ -149,6 +156,7 @@ public class AssetsController implements Initializable {
         addAssetBtn.setOnMouseClicked(mouseEvent -> uiUtilities.changeScene(rulTimeline, TextConstants.ADD_ASSETS_SCENE, addAssetBtn.getScene()));
 
         sortingSetUp();
+        searchAssets();
     }
 
     private void sortingSetUp() {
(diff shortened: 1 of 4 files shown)


JavaFXUserInterface/src/main/java/controllers/AssetsController.java
```diff
@@ -168,18 +176,33 @@ private void sortingSetUp() {
                 switch (newValue) {
                     case SORT_DEFAULT:
                         FXCollections.sort(assets, Comparator.comparingInt(Item::getId));
+                        if (!search.getText().trim().isEmpty()) {
```
JayChouteau: Having the conditional statement every time for every switch case is a little redundant. Maybe you could use the setAssetToDisplay() method to establish a local assets list for the sort asset button to use. You would establish that local list before the switch statement. This way you could remove all the added conditional statements and keep the initial switch case with the local list of assets to display instead of the class member assets

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
```diff
@@ -190,99 +213,162 @@ private void sortingSetUp() {
...
+                    searchMatch = "Match";
+                }
+            } else {
+                searchMatch = "No Search";
+            }
+
+            if (assetsTabPane.getSelectionModel().getSelectedItem().getId().equals("thumbnailTab")) {
+                assetsThumbPane.getChildren().clear();
+                generateThumbnails();
+            } else {
```
JayChouteau: This is really cherry-picking, but I think that having else if with the list tab specified instead of else might provide more scalability in case there are future uses and addition of other tabs. It also clarifies for developers the intention of this conditional statement (that the following code is meant for a specific tab as well, not for all other possibilities). You can let me know what you think.

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
```diff
@@ -190,99 +213,162 @@ private void sortingSetUp() {
...
+     * When a search has no match, all system panes are cleared effectively showing no assets.
+     *
+     * @author Najim
+     */
+    private void searchAssets() {
+        searchedAssets = FXCollections.observableArrayList();
+        search.textProperty().addListener((observable, oldValue, newValue) -> {
+            searchedAssets.clear();
+            if (!newValue.equals("")) {
+                assets.stream().filter(o -> StringUtils.containsIgnoreCase(o.getRul().get(), newValue) || StringUtils.containsIgnoreCase(o.getName(), newValue)
```
JayChouteau: I was simply wondering if it wouldn't have been easier and more readable to just iterate through each asset's member variables and check the StringUtils.containsIgnoreCase(...) since the calling function is used and reused. It might be more complicated as well. I am just asking.

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
```diff
@@ -190,99 +213,162 @@ private void sortingSetUp() {
...
      * Creates elements that are in the scene so the data can be displayed.
      *
      * @author Jeff
      */
     public void generateThumbnails() {
         ObservableList<Pane> boxes = FXCollections.observableArrayList();
+        ObservableList<Asset> assetsDisplayed = setAssetListToDisplay();
 
-        for (Asset asset : assets) {
+        if (assetsDisplayed != null) {
```
JayChouteau, line 297: I think it would be good to provide an alternative text to display on the page in the case where there are no assets to display. What I had in mind would be kind of a grey text appearing in the middle of the screen stating that there are "No results for your search" or something like that.

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
```diff
@@ -342,14 +429,16 @@ public void generateList() {
...
 
-        table.setItems(assets);
-        table.setId("listTable");
-        table.getColumns().addAll(assetTypeCol, serialNoCol, modelCol, modelRULCol, recommendationCol, locationCol, siteCol, categoryCol, manufacturerCol, descriptionCol);
-        AnchorPane.setBottomAnchor(table, 0.0);
-        AnchorPane.setTopAnchor(table, 5.0);
-        AnchorPane.setRightAnchor(table, 0.0);
-        AnchorPane.setLeftAnchor(table, 0.0);
-        assetsListPane.getChildren().addAll(table);
+        if (assetsDisplayed != null) {
```
JayChouteau: I think it would be good to provide an alternative text to display on the page in the case where there are no assets to display. What I had in mind would be kind of a grey text appearing in the middle of the screen stating that there are "No results for your search" or something like that.

JavaFXUserInterface/src/test/java/UnitTests/Controllers/AssetsControllerTest.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        FxToolkit.hideStage();
+        release(new KeyCode[]{});
+        release(new MouseButton[]{});
+    }
+
+    @Test
```
JayChouteau, line 43: These tests are great tests, but they are more system tests than unit tests. At least that's how I feel, but maybe we could discuss about it at the meeting or take a few minutes to discuss it.

JayChouteau (changes requested):
Good work on the user story. The logic is good and all acceptance criteria have been met. Solid work overall. I left a few comments. Some of them are request changes, but others are more discussions. In any case, I think it would be great if you could answer them all.

JavaFXUserInterface/src/test/java/UnitTests/Controllers/AssetsControllerTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private AssetsController assetController;
+    private Scene scene;
+
+    @Override
+    public void start(Stage stage) throws Exception {
+        FXMLLoader fxmlLoader = new FXMLLoader(AssetsController.class.getResource("/Assets.fxml"));
+        Parent root = fxmlLoader.load();
+        scene = new Scene(root);
+        assetController = (AssetsController) fxmlLoader.getController();
```
JayChouteau: IntelliJ warning: the casting to (AssetsController) is redundant and be removed.

JavaFXUserInterface/src/test/java/UnitTests/Controllers/AssetsControllerTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertTrue(assetController.setAssetListToDisplay().size() > 0);
+        assertTrue(rootNode.getChildren().size() > 0);
+    }
+
+    @Test
+    public void invalidSearchTest() {
+        clickOn("#search").write("This is an invalid search");
+        FlowPane rootNode = (FlowPane) scene.getRoot().lookup("#assetsThumbPane");
+        verifyThat("#search", TextInputControlMatchers.hasText("This is an invalid search"));
+        interact(() -> {
```
JayChouteau: IntelliJ warning: This can be changed to a lambda expression.

JayChouteau (changes requested):

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
najimgh: I tried this solution at first. However, I was not able to make it work. For some reason, the list of assets would not get generated again even though the code did execute generateThumbnails() properly.

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
najimgh: Great suggestion. The way you explained it makes more sense to use else if

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
najimgh: Doing it this way or iterating through each asset's member variable would not change much except for readability and understanding of the code. Although, Iterating through the assets might result in more lines of code. 
I have added comments to explain what asset.stream().filter() does

JavaFXUserInterface/src/test/java/UnitTests/Controllers/AssetsControllerTest.java
najimgh: Fixed

JavaFXUserInterface/src/test/java/UnitTests/Controllers/AssetsControllerTest.java
najimgh: Fixed

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
JayChouteau: Yeah, I was talking about readability and understanding of the code. But I wasn't sure what was more readable between the stream and for/if-else statements. This way is fine.

TalalBazerbachi (changes requested):
I don't know if it's working as intended.
- I tried to search by recommendation 'ok' but it also included a failed asset.
image

- When I typed '101', it gave more than one asset:
image

-Also why is the performance of the search bar slow?

najimgh:
@t98b  
This is a normal behaviour of the search functionality. We compare the value of each member variable of the assets. Therefore it's normal to have multiple results for '101' since 101 can be found in RUL, Serial number, etc. To refine the search, we can order and filter the list of assets.

The performance is slow only when typing the first 2 or 3 characters as it is looking through the 704 assets available. Moreover, since this is a live search, we are generating and clearing the Thumbnails every time a character is typed. The only way I see that we could improve the performance would be to not make it a live search.

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
najimgh: The solution we talked about did not work, I tried some other things without any success. I think the way it is now is the better option so far.

JayChouteau: That is fine then. Leave it this way for now. WE will see when we implement more filters and sorting options.

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
najimgh: Table now shows "No results found"

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
najimgh: "No results found" is now displayed if no match has been found from the search

najimgh:
@t98b 
Improved performance of the live search functionality by delaying the search of assets (delay of 0.25 seconds) in case the user keeps typing.

JayChouteau (approved):
Nice job on improving the search performance by adding a delay in listening to characters. Also, nice job of indicating where no results were found and well done on handling null values for the sorting. The only comment I would make is that I feel like it is weird that the results not found for the list and results not found for the thumbnails are different. Other than that, pretty good job.

JavaFXUserInterface/src/test/java/UnitTests/Controllers/AssetsControllerTest.java
JayChouteau: We can leave this as is, I consider this issue done

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
JayChouteau: Nice touch on the no results found.

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
JayChouteau: Well done. The only comment I would make is that it feels weird that the display of "No results found" is different for the list view than the thumbnail view. Was there a specific reason for it? Is it because of the flow pane? Otherwise, I would have left it the same as the list view.

TalalBazerbachi (approved):

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
najimgh: Yes, because of the flow pane I was not able to make it the same as the list view

JavaFXUserInterface/src/test/java/UnitTests/Controllers/AssetsControllerTest.java
najimgh: I couldn't find any relevant information for unit testing UI and more specifically the search bar

PR #309, opened 2021-03-26
248 images

### LINKED ISSUE #248: UI Feature - Images
As a user, I would like to see images for the assets that I have.

Description
On the asset information page, the user would like to have the image of that specific asset displayed among the other info. There is currently uncertainty as to whether or not adding images for assets can be done dynamically. The first part would be to explore the possibilities.

Acceptance Criteria:
- [x] There is an image of the asset on that asset information page
- [x] Display image in Thumbnail on Assets page
- [x] Allow user to select image on Add Asset page
- [ ] Allow user to edit image on Edit Asset page [REMOVED]
- [x] If the asset does not have an image, a default display will be shown instead
- [x] Only the minimum number of images need to be added

rdsaliba:
Changes summary: 
- There is an image of the asset on that asset information page
- An image is displayed in thumbnail on Assets page
- User can upload an image on Add Asset page and associate it with the new asset
- If the asset does not have an image, a default image is shown
- Only the minimum number of images need to be added

In order to test the feature, run the sql script added in the pull request.

Expected behaviour at run time:
- All assets have a default image

One way of testing would be to create 3 new assets with a picture with similar format to default.png.
Look in the table "picture" to check new entries.
Use the imageId of the picture recently inserted and set it to an asset with similar column name.


TalalBazerbachi (dismissed):
I tried returning from asset type page to assets page but I got an error:
image

Khal-Shah:
Was version 19 built on top of the 18 here? I'm having these errors on 19:

image

rdsaliba:
> Was version 19 built on top of the 18 here? I'm having these errors on 19:
> 
> image

Hi @Khal-Shah, you did not run the script provided in the pull request

Server/src/main/java/local/AssetDAOImpl.java
```diff
@@ -228,6 +228,7 @@ public Asset createAssetFromQueryResult(ResultSet assetsQuery) throws SQLExcepti
         newAsset.setSerialNo(assetsQuery.getString("sn"));
         newAsset.setRecommendation(assetsQuery.getString("recommendation"));
         newAsset.setAssetTypeName(assetsQuery.getString("asset_type.name"));
+        newAsset.setImageId(assetsQuery.getInt("imageId"));
```
JayChouteau, line 232: This is added in this creation of asset method, but was not added in the other asset creation method on line 244. I know some of it is outside of the scope of this issue, but it seems like the 2 methods are doing the exact same thing. I think it would be good to delete one of the 2 and replace it with the other one where it was used. This way, you also only have to add the setImageId line at one specific place

JavaFXUserInterface/src/main/resources/AddAsset.fxml
```diff
@@ -107,6 +107,8 @@
                     <Font name="Segoe UI" size="14.0"/>
                 </font>
             </Text>
+         <Button fx:id="uploadBtn" layoutX="650.0" layoutY="97.0" mnemonicParsing="false" text="Upload image" styleClass="uploadBtn"/>
```
JayChouteau: The upload button seems to be overlapping with the input validation messages. It would be a good idea to move it farther up to the side. 
inputvalidation

JavaFXUserInterface/src/main/java/external/AssetDAOImpl.java
```diff
@@ -64,12 +69,57 @@ public void insertAsset(Asset asset) {
...
             ps.setString(7, asset.getSite());
             ps.setString(8, asset.getLocation());
+            ps.setInt(9, asset.getImageId());
             ps.executeQuery();
         } catch (SQLException e) {
             logger.error("Exception in insertAsset(): ", e);
         }
     }
 
+    public PreparedStatement storeImage(FileInputStream fileInputStream, String name) {
```
JayChouteau, line 88: Missing @override tag to put in the interface + method description

JavaFXUserInterface/src/main/java/external/AssetDAOImpl.java
```diff
@@ -64,12 +69,57 @@ public void insertAsset(Asset asset) {
...
+            ps.setString(2, name);
+            ps.executeQuery();
+        } catch (SQLException | IOException e) {
+            logger.error("Exception in storeImage: ",  e);
+        }
+
+        return ps;
+    }
+
+    public Image findImageById(int imageId) {
```
JayChouteau, line 109: Missing @override tag to put in the interface + method description

JavaFXUserInterface/src/main/java/external/AssetDAOImpl.java
```diff
@@ -64,12 +69,57 @@ public void insertAsset(Asset asset) {
...
+                InputStream inputStream = blob.getBinaryStream();
+                image = new Image(inputStream);
+            }
+        } catch (SQLException e) {
+            logger.error("Exception in findImageById(): ", e);
+        }
+        return image;
+    }
+
+    public int findImageIdByName(String name) {
```
JayChouteau, line 133: Missing @override tag to put in the interface + method description

JavaFXUserInterface/src/main/java/controllers/AddAssetController.java
```diff
@@ -53,11 +63,21 @@
     @FXML
     private TextField locationInput;
     @FXML
+    private AnchorPane inputError;
```
JayChouteau: This AnchorPane does not exist anymore and needs to be removed.

JayChouteau (dismissed):
Really nice job on the issue!! It is clean and well done. The heap memory increase allows us to see all assets with images. It is too bad we can't really test adding assets with images due to the bug associated with handling new assets, but that will be fixed later. The hovering of buttons is a nice addition that should be added to the whole project! I left a few comments on the code that needs attention.

Server/src/main/java/local/AssetDAOImpl.java
rdsaliba: Great catch, I did not see this one. I'll remove it in my next commit.

JayChouteau (approved):
Thanks for addressing my comments! Approved!

Khal-Shah (approved):
Because of the add asset bug we currently have, the image feature cannot be fully verified through the UI by uploading a new image for a new asset and associating that image id to an already existing asset. But it can be checked in the database that newly added asset has the correct image id. 

We will need to test this feature again once the add asset bug has been fixed but the implementation for this issue seems well done and meets the acceptance criteria. Though the performance in the asset thumbnail page seems to be lagging a bit compared to before, the proposed solution to apply cache to panes should improve it. 

Approved.

PR #320, opened 2021-04-04
#137 added filtering functionality

### LINKED ISSUE #137: Add a filter for asset types and status on the "assets" view
As a user, I want to be able to choose which assets I want to look at depending on the chosen asset type.

Acceptance Criteria:
- [x] Mock-up documentation update on Confluence (link to be added to this issue in comments)
- [x] Have a filter for asset types and status on the assets view, both list and thumbnails view
- [x] The filter shows existing asset types
- [x] The filter shows assets for a certain status
- [x] The filter allows multiple answers entries
- [x] Once a or multiple asset types have been chosen, only those will show on both the list view and thumbnails view
- [x] Once a or multiple status have been chosen, only those will show on both the list view and thumbnails view
- [ ] Unit tests

PaulMicu:
this resolves #137


JavaFXUserInterface/src/main/java/controllers/AssetsController.java
```diff
@@ -171,9 +177,115 @@ public void attachEvents() {
...
+        int i = 0;
+        CheckBox option;
+        for (AssetType assetType : assetTypeDAO.getAssetTypeList()) {
+            assetTypeFilterCondition.put(assetType.getName(), false);
+            option = new CheckBox(assetType.getName());
+            option.selectedProperty().addListener((observable, oldValue, newValue) -> {
+                assetTypeFilterCondition.replace(assetType.getName(), newValue);
+                generateContent();
+            });
+            navList.add(option, 0, i);
```
JayChouteau, line 227: The use of int i could be replaced by automatic call to the index in the list, such as : 
assetTypeDAO.getAssetTypeList().indexOf(assetType)

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
```diff
@@ -171,9 +177,115 @@ public void attachEvents() {
...
+        }
+        i = 0;
+        for (ThresholdEnum thresholdEnum : ThresholdEnum.values()) {
+            thresholdFilterCondition.put(thresholdEnum.getValue(), false);
+            option = new CheckBox(thresholdEnum.getValue());
+            option.selectedProperty().addListener((observable, oldValue, newValue) -> {
+                thresholdFilterCondition.replace(thresholdEnum.getValue(), newValue);
+                generateContent();
+            });
+            navList.add(option, 1, i);
```
JayChouteau: The use of int i could be replaced by automatic call to the index in the enum values, such as: 
ThresholdEnum.valueOf(thresholdEnum.getValue()).ordinal()

JayChouteau (dismissed):
Logic and code are clean. All the acceptance criteria have been met and the filter addition was well integrated with the search bar and the generate thumbnails and generate list. Good re-use of code. The solution for the filter being a button and nav list works well with out app and the slide in transition adds to the smoothness of the UI. I left 2 small comments on code snippets you can reply on and let me know what you think. The other 2 comments I have would be those:
1) I feel like it would be a good idea to have a column header on top of each filter type on the nav list, such as "Asset type" and "Threshold" to guide the user
2) This could be part of Jeff's issue, but hovering highlight for the filter button would be a nice add

rdsaliba (approved):
Hey Paul, in my case I don't see any issue with the way it is implemented. Works perfectly and there does not seem to be a bug associated with the changes. I however agree with Jeremie for the uniformity of the buttons and all, it will need to be addressed hopefully before the end of the project!

najimgh (approved):
Acceptance Criteria have been met. The filtering works well with the ordering and searching of assets. Nice touch with the animation when clicking on the Filter button.

JavaFXUserInterface/src/main/java/controllers/AssetsController.java
PaulMicu: it would create too many db calls for nothing, i could do a traditional for i loop but i dont see any benefits

PR #321, opened 2021-04-04
#138 allows for an asset to be edited

### LINKED ISSUE #138: UI Feature - Editing an asset
As a user, I would like to be able to edit an existing asset and modify the information linked to that asset.

Additional Information:
Put "Modify" or "Edit" button on Asset Info page

Acceptance criteria:
- [ ] Create the mockup and publish it on Confluence (link to be inserted in this issue)
- [ ] edit button on the System info view
- [ ] Once clicked, the edit button brings the user to a page where he can edit the system info
- [ ] The modifications can either be saved or canceled
- [ ] When canceled, a confirm dialog asks the user about his action
- [ ] Once saved, the app brings back to the system info page with updated info
- [ ] If canceled, the app brings back to the system info page with previous info
- [ ] Input validation
- [ ] Edit everything except for SerialNo and Asset Type
- [ ] unit tests

PaulMicu:
Slight change from the initial solution, change and save is done directly in the Asset Info page instead of going to a different page, similar to what we do in edit asset type
this resolves #138


JavaFXUserInterface/src/main/resources/css/revamp/assetInfo.css
```diff
@@ -122,4 +122,11 @@
     -fx-font-family: "Segoe UI Semibold";
     -fx-font-size: 20px;
     -fx-padding: 5 15 5 15
-}
\ No newline at end of file
+}
+#assetSaveBtn {
```
JayChouteau, line 126: This save Button doesn't seem to be fitting with the rest of the styling used for the other save buttons on other screens. The color, pref-width, and height as well as other styling are different. See add asset and add asset type for example

JavaFXUserInterface/src/main/java/utilities/FormInputValidation.java
```diff
@@ -199,6 +199,41 @@ public static boolean assetFormInputValidation(AnchorPane textInputControlAnchor
...
+        assetValidForm = true;
+        ArrayList<TextInputControl> textInputControls = new ArrayList<>();
+
+        // Initializing the form input validation
+        initializeFormInputValidation(textInputControlAnchorPane, textInputControls);
+
+        logger.info("Start - formInputValidation() -> The form is : {}", assetValidForm);
+
+        // Input validation on the different asset input fields
+        assetDescriptionValidation(textInputControlAnchorPane, assetDescriptionInput, textInputControls.indexOf(assetDescriptionInput));
```
JayChouteau, line 228: There is no validation on the name even though you can edit the asset name. This causes some issues when you enter a name that is too long. And the user can't know because he is brought back to the assets thumbnail screen. He will only notice if he comes back on the asset info page.
asset_name_1

JavaFXUserInterface/src/main/java/controllers/AssetInfoController.java
```diff
@@ -231,6 +246,17 @@ public void attachEvents() {
             rawDataListPane.getChildren().clear();
             generateRawDataTable();
         });
+
+        assetSaveBtn.setOnMouseClicked(mouseEvent -> {
```
JayChouteau, line 275: The save button could be disabled as long as any of the textfield hasn't been edited, to keep it consistent with other save buttons in our app.

JayChouteau (changes requested):
The editing directly on the asset info page works well with our application. The transformation from strings to a simpleStringProperty also makes it easier to keep track and update the values. I left a few comments in the code. Other than that, I have 2 comments:
1) The input validation seems to be overlapping with some textfields. Given the extra white space we have to the right, we could move the second column to the right so the input validation messages dont overlap on labels.
errors_asset

2) I think it would be good to have prompt texts on the text fields to appear if the fields are empty. You can let me know what you think of this.

JayChouteau:
The acceptance criteria of this issue should be modified and checked if done.

JavaFXUserInterface/src/main/java/controllers/AssetInfoController.java
rdsaliba: I agree with the point, simply because it has to be the same across the app.

JavaFXUserInterface/src/main/java/utilities/FormInputValidation.java
rdsaliba: Good catch

rdsaliba (commented):
Hi Paul, I think Jeremie said it all. I'm ok with the feature and the code, just need to correct the buttons and validation on the form!

JavaFXUserInterface/src/main/resources/AssetInfo.fxml
```diff
@@ -1,19 +1,14 @@
...
+<?import javafx.scene.control.*?>
+<?import javafx.scene.image.*?>
+<?import javafx.scene.layout.*?>
+<?import javafx.scene.text.*?>
 <?import javafx.scene.layout.AnchorPane?>
 <?import javafx.scene.layout.FlowPane?>
-<?import javafx.scene.text.Font?>
-<?import javafx.scene.text.Text?>
 
-<AnchorPane maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" prefHeight="834.0" prefWidth="1194.0" xmlns="http://javafx.com/javafx/15.0.1" xmlns:fx="http://javafx.com/fxml/1" fx:controller="controllers.AssetInfoController">
```
najimgh, line 16: The AnchorPane has been removed and so has the association with the controller. I believe this is what is causing the error mentioned above

najimgh (dismissed):
I get the following error when I try to click on any asset to see its information

Screen Shot 2021-04-08 at 12 24 14 AM

JavaFXUserInterface/src/main/resources/AssetInfo.fxml
PaulMicu: yep looks like an error during merge conflict resolution

rdsaliba (approved):
Good job.

JayChouteau (approved):
Thanks for addressing the changes. Approved.

